PR #8, opened 2023-08-04
Cleanup: Add getRecipientAddress to RelayerAbstract

barnjamin:
Every context had some check to see if the transfer was meant for Solana since it requires a different receiver address than the main account.

This pr adds a `getRecipientAddress` to the RelayerAbstract class with the default implementation of returning the same address. Currently only `SolanaContext` overrides this method to provide the ATA address

### DIFF
--- a/packages/evm/src/context.ts
+++ b/packages/evm/src/context.ts
@@ -27,8 +27,6 @@ import {
   parseVaa,
   RelayerAbstract,
   createNonce,
-  MAINNET_CHAINS,
-  Network,
 } from '@wormhole-foundation/connect-sdk';
 import { EvmContracts } from './contracts';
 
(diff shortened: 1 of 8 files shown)


packages/evm/src/context.ts
```diff
@@ -190,22 +190,18 @@ export class EvmContext extends RelayerAbstract<ethers.ContractReceipt> {
...
-        recipientAddress,
-      );
-    }
+    let tokenID: TokenId =
+      token === NATIVE
+        ? {
+            address: await bridge.WETH(),
+            chain: sendingChainName,
+          }
+        : token;
```
barnjamin, line 199: should be const but also, this method of `native` to `TokenId` should be a util fn on the class if we need it in >1 spot

kev1n-peters (approved):

PR #9, opened 2023-08-04
Docs: adding typedoc

barnjamin:
Adding docs using the built-in package support to document all the subpackages into a single directory we can serve from GH pages


packages/solana/src/utils/wormhole/coder/idl.ts
```diff
@@ -54,12 +54,6 @@ export class IdlCoder {
       case 'i128': {
         return borsh.i128(fieldName);
       }
-      case 'u256': {
```
kev1n-peters, line 57: Why are we removing these? Are they not used anywhere?

packages/examples/dist/src/index.js
kev1n-peters, line 1: I don't think we should add the files in the `dist` directory

packages/solana/src/utils/wormhole/coder/idl.ts
barnjamin: I don't think they're used and they puke when I try to build docs.

See https://github.com/nonergodic/sdkv2/pull/3#discussion_r1268194985

packages/examples/dist/src/index.js
barnjamin: doh! removing

kev1n-peters (approved):

PR #14, opened 2023-08-08
Add a method to get raw vaa in addition to the parsed one

barnjamin:
Some methods require the raw vaa as Uint8Array or Buffer, besides providing a method to return the parsed VAA we should provide the raw VAA as well

### DIFF
--- a/packages/sdk-core/src/wormhole.ts
+++ b/packages/sdk-core/src/wormhole.ts
@@ -21,7 +21,7 @@ import {
   TokenId,
   WormholeConfig,
 } from './types';
-import { ParsedVaa, parseVaa } from './utils/vaa';
+import { ParsedVaa, SignedVaa, parseVaa } from './utils/vaa';
 import { SolanaAbstract } from './abstracts/contexts/solana';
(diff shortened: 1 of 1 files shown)


packages/sdk-core/src/wormhole.ts
```diff
@@ -417,17 +417,30 @@ export class Wormhole extends MultiProvider<Domain> {
...
+  }
+
+  /**
+   * Gets a VAA from the API or Guardian RPC and parses it, finality must be met before the VAA will be available.
+   *  See {@link ChainConfig.finalityThreshold | finalityThreshold} on {@link MAINNET_CONFIG | the config}
+   *
+   * @param msg The MessageIdentifier used to fetch the VAA
+   * @returns The Parsed VAA if available
+   */
+  async getVAA(msg: MessageIdentifier): Promise<ParsedVaa | undefined> {
```
anondev2323, line 440: How about calling this `getParsedVAA`? And the one above just `getVAA` (but `getVAABytes` is ok too).

anondev2323 (approved):

PR #16, opened 2023-08-09
Combine Repos

barnjamin:


core/base/src/constants/finality.ts
```diff
@@ -0,0 +1,54 @@
...
+] as const satisfies readonly (readonly [
+  Network,
+  readonly (readonly [ChainName, number])[]
+])[];
+
+const mapping = {
+  Mainnet: toMapping(finalityThresholds[0][1]),
+  Testnet: toMapping(finalityThresholds[1][1]),
+  // TODO: reusing testnet finality, ok?
+  Devnet: toMapping(finalityThresholds[1][1]),
```
anondev2323, line 47: I've been a little bit confused by Testnet and Devnet tbh... I'm wondering if we need a Devnet mapping or if we just have Mainnet and non-Mainnet? As far as I know, we usually pick one or the other, but we don't have any deployed to both.  And it's not super clear which chains are deployed on Testnet vs Devnet.  I'm not really recommending anything, just airing my grievances

core/base/src/constants/modules.ts
```diff
@@ -0,0 +1,39 @@
...
+ *   but ["Core", "TokenBridge", "NFTBridge", "WormholeRelayer"] seems to be more consistent given
+ *   current naming "conventions" 
+ */
+
+export const modules = [
+  "CoreBridge",
+  "TokenBridge",
+  "NftBridge",
+  "Relayer",
+//TODO "Connect"
```
anondev2323, line 34: I wouldn't list Connect as a module, but rather a consumer of all of the above

core/base/src/constants/modules.ts
```diff
@@ -0,0 +1,39 @@
...
+ * My preferred choice would be ["CoreBridge", "TokenBridge", "NftBridge", "Relayer"]
+ *   but ["Core", "TokenBridge", "NFTBridge", "WormholeRelayer"] seems to be more consistent given
+ *   current naming "conventions" 
+ */
+
+export const modules = [
+  "CoreBridge",
+  "TokenBridge",
+  "NftBridge",
+  "Relayer",
```
anondev2323, line 33: Probably good to rename to TokenBridgeRelayer, to not confuse with generic relayer

core/base/src/constants/platforms.ts
```diff
@@ -0,0 +1,119 @@
+import { ChainName } from "./chains";
+import {
+  column,
+  toMapping,
+  toMappingFunc,
+  reverseArrayMapping,
+} from "../utils/mapping";
+
+const platformAndChainsEntries = [
```
anondev2323, line 9: Oof, need to make terminology consistent across repo

package.json
```diff
@@ -22,14 +22,13 @@
...
-    "packages/sdk-core",
-    "packages/aptos",
-    "packages/evm",
-    "packages/solana",
-    "packages/sei",
-    "packages/sui"
+    "core/base",
+    "core/definitions",
+    "core/connect",
+    "platforms/evm"
```
anondev2323, line 32: Missing the other platforms?

barnjamin: others wont build yet with new backend stuff

core/base/src/constants/modules.ts
```diff
@@ -0,0 +1,39 @@
...
+ * 
+ * Additionally, "modules" seems like a bad choice in a TS/JS context...
+ * 
+ * With all of this in mind: What should we name modules here?
+ * My preferred choice would be ["CoreBridge", "TokenBridge", "NftBridge", "Relayer"]
+ *   but ["Core", "TokenBridge", "NFTBridge", "WormholeRelayer"] seems to be more consistent given
+ *   current naming "conventions" 
+ */
+
+export const modules = [
```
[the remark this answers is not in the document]
barnjamin: I'm not tied to it but it is the name used in at least a few places: https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0002_governance_messaging.md#general-packet-structure

core/base/src/constants/platforms.ts
barnjamin: sorry which term? platform?

README.md
```diff
@@ -9,16 +9,16 @@ A developer would use the core connect-sdk package in conjunction with 1 or more
 Getting started is simple, just import and pass in the contexts to the Wormhole class.
 
 ```ts
-import { Wormhole, Context, Network } from '@wormhole-foundation/connect-sdk';
+import { Wormhole, Context } from '@wormhole-foundation/connect-sdk';
 import { EvmContext } from '@wormhole-foundation/connect-sdk-evm';
 import { SolanaContext } from '@wormhole-foundation/connect-sdk-solana';
 
-const NETWORK = Network.MAINNET;
+const network = "Mainnet";
```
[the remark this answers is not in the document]
anondev2323: I def prefer the const.  Because you can mess up a string pretty easily, like it can be hard to remember if its "Mainnet", "mainnet" or "MAINNET"

core/base/src/constants/platforms.ts
anondev2323: Yes, Platform.  I think Andreas has been using that term which is ok with me.  But I had been using the term `Context`, and I know there's been other instances of us using different terminology for the same things

core/base/src/constants/modules.ts
nonergodic: Connect to me is the token bridge relayer which fits the mold of a module:
It has its own set of contracts, payloads, etc.

README.md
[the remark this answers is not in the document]
nonergodic: Imho the main (and really only) advantage of the const is that it stands out as being a const over a normal string.

However, I think that the disadvantage of the consts strongly outweigh that:

* Definitions become more awkward:
Take the example of networks:
`export const networks = ["Mainnet", "Testnet", "Devnet"] as const;
export type Network = (typeof networks)[number];
export const isNetwork = (network: string): network is Network =>
  networks.includes(network as Network);`
If you want to declare consts you will have to repeat yourself in so way:
either:
`const MAINNET = "Mainnet";
const TESTNET = "Testnet";
const DEVNET = "Devnet";
const networks = [MAINNET, TESTNET, DEVNET];`

or the other way around by using a destructuring assignment.

And this is just the tip of the iceberg:
Do we introduce constants for all chain names?
If yes: Is SOLANA now defined in chains or in platforms?
If no: That's oddly inconsistent.

* Whenever a type is expected you see the definition of the type e.g. if you have a function `doSomething(network: Network)` your editor will tell you that the 3 valid values are the 3 string literals. But if we have constants, instead of using the straight forward string literals, you are supposed to import and use a constant instead even though nothing in the code tells you to. So even if we are conscientious with using the constants in the sdk itself you're all but guaranteed to have user code that will simply use the string literals anyway.

* There is no risk of typos or spelling confusion since the Typescript will enforce that you are passing a correct type literal and instead of being all capitalized it's simply pascal case which is easier to type anyway (especially if - like me - you have no capslock on your keyboard)

core/base/src/constants/modules.ts
nonergodic: Relayer is the generic Relayer atm :sweat_smile:

core/base/src/constants/platforms.ts
nonergodic: 'Context' is too much of a generic term in programming to use for platforms (which is also questionable).
Imho it's similar to 'environment' where it's natural to jump to 'environment variable' etc.

nonergodic: Maybe ChainFamily would be better because it is less generic and also reasonable descriptive...

but since Ben and I agreed that we'd rename Chain, Platform, Module, ... to ChainName, PlatformName, ModuleName, this would then turn it into ChainFamilyName... :grimacing:

nonergodic: (and `Module` is a terrible name for the same reason - we should really come up with something better if we can)

core/base/src/constants/modules.ts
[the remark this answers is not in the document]
nonergodic: agree with @anondev2323 - it has the current name for the reasons that Ben points out...

I don't like application or xApp because those names should be reserved for what users build with the sdk.. (and CoreBridge really isn't an application - it's effectively the operating system...)

README.md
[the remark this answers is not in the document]
anondev2323: Ok yeah, good points, let's go with the strings 👌

core/base/src/constants/modules.ts
nonergodic: beyond the length, it's a bit ugly to have one module name also be the prefix of another module name... :confused: 

It's certainly much more descriptive though. I guess my thought here is: It's the module that powers the connect widget. I agree that in light of the name of the sdk "connect" is a disgustingly bad term - but imho it's the SDK who is at fault here <insert principal Skinner meme here>.

nonergodic: In a similar vein, "Relayer" is pretty unspecific, but imho still better than what seems to be its canonical name atm i.e. "WormholeRelayer" (because in an SDK by Wormhole whose Relayer will it be?) or the alternative of "AutomaticRelayer" (as opposed to what sort of relayer? a non-automatic one? as in no relayer whatsoever?)

I feel like were being backed into a corner her by all the other terrible naming choices that have been made (a bit of a theme I suppose).

README.md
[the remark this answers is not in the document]
barnjamin: same

core/base/src/constants/modules.ts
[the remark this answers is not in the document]
barnjamin: How about `Protocol`? 

And further specify the protocols described here https://github.com/wormhole-foundation/wormhole/tree/main/whitepapers

PR #19, opened 2023-08-14
Fix/build package

bradleySuira:
This Pull request was created because we were not able to  current version of @wormhole-foundation/connect-sdk  (https://www.npmjs.com/package/@wormhole-foundation/connect-sdk) inside a dapp because of Javascript module system configuration and build tasks.

The solution:
Create a Hybrid npm package that has support for  ESM and CommonJS at the same time so it can be used in apps with that has support for new ES modules or existing commonJs. We consider ESM the future, but want to continue to support CommonJS users

The changes:
- update imports to specify the extension .js
- create two separated tsconfig.json files one for esm and another for commonJs
- update some parts of the code to fix typings usage
- update npm tasks and package.json to support the two different configs at the same time
- update jest settings


barnjamin:
Hey @bradleySuira thanks for the PR!

I think we can do this without adding `.js` extension to all the imports, I've got this implemented in this branch: https://github.com/wormhole-foundation/connect-sdk/pull/16

The API for that branch will be v different from the current one so I wont close this PR until that one is merged in, just in case others want this same behavior with the current interface

bradleySuira:
Thanks @barnjamin

PR #26, opened 2023-08-22
New SDK Interfaces

barnjamin:


DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+- Platform Context
+    A class which implements a standardized format and set of methods. It may include additional chain-specific methods and logic.
+- Protocol 
+    A cross-chain application built on top of Wormhole (the core contracts are also considered a module)
+- Universal Address
+    A 32-byte address, used by the wormhole contracts
+- Native Address (I think this should be called "Platform Address")
+    An address in the standard chain-specific format
+- Native
+    The "home" chain (e.g. ETH is native to Ethereum)
```
nonergodic, line 137: good point/agree re Platform Address

I guess if we want to use the word "native", we should have some coherent idea what it should represent, because right now it's quite the confusing grab-bag of different concepts (e.g. calling the gas token of a chain the native token, calling the origin token of the token bridge the native / home chain etc.)

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+## Can we provide some way to make other non-standard applications available to use through the WormholeTransfer?
+
+Say I have an app that defines its own protocol, can I provide something that adheres to the WormholeTransfer interface so a dev can install it and call it like the TokenTransfer?
+
+
+# Outstanding Questions: 
+
+What is the preferred terminology to refer to either end of a cross-chain message: from/to, source/target or origin/destination?
+
+What is the preferred terminology for the core Wormhole layer? (i.e. Core Contracts or Wormhole Contracts)
```
nonergodic, line 189: I feel like we've (partially) had this discussion before, but I can't quite remember where/in what context. :thinking:

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+- It _may_ be backed by a wallet but not necessarily, as long as it fulfils the interface
+
+## Can we provide some way to make other non-standard applications available to use through the WormholeTransfer?
+
+Say I have an app that defines its own protocol, can I provide something that adheres to the WormholeTransfer interface so a dev can install it and call it like the TokenTransfer?
+
+
+# Outstanding Questions: 
+
+What is the preferred terminology to refer to either end of a cross-chain message: from/to, source/target or origin/destination?
```
nonergodic, line 187: I feel like we've (partially) had this discussion before, but I can't quite remember where/in what context. 🤔

DESIGN.md
nonergodic, line 189: Currently the protocol name is CoreBridge and primarily refers to the contract and its associated payloads. Core layer in my mind should include the entire attestation mechanism (i.e. including off-chain components like guardians etc.).

Bridge is perhaps not the best name because in its raw form it really is just attestation which, if anything, is a broadcast.

I was reluctant to call it WormholeCore or anything else with Wormhole in the name because we're already in a Wormhole SDK and just "Core" seemed too generic.

Open to suggestions. Main primary concern is consistency.

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+
+    e.g.  
+    // need to implement universal univeral decoder?
+    given eth address, and without installing evm platform, how do i turn it into a solana wrapped token without knowing how to fmt the address? 
+
+    // neet to tweak contracts on sol to support !ATA?
+    given an xfer from eth=>sol, and without installing sol platform, how do i determine the ATA?
+
+
+What is the benefit of costmap vs single fat object
```
nonergodic, line 203: primarily: easily supporting multiple "views" (e.g. inverse mappings)
also: proper bigint key support (bigint is not a `PropertyKey`)

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+Say I have an app that defines its own protocol, can I provide something that adheres to the WormholeTransfer interface so a dev can install it and call it like the TokenTransfer?
+
+
+# Outstanding Questions: 
+
+What is the preferred terminology to refer to either end of a cross-chain message: from/to, source/target or origin/destination?
+
+What is the preferred terminology for the core Wormhole layer? (i.e. Core Contracts or Wormhole Contracts)
+
+Should we namespace export base/definitions?
```
nonergodic, line 191: We shouldn't export the payloads by default because they only clutter up the namespace unnecessarily. E.g. users don't care about governance payloads and so they shouldn't be added to the payload factory by default.

Being a C++ dinosaur and fan-boy I think we should follow the same principles "you don't pay for things you don't use".

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+Should we namespace export base/definitions? 
+
+How should we think about xchain concepts without having xchain context?
+
+    e.g.  
+    // need to implement universal univeral decoder?
+    given eth address, and without installing evm platform, how do i turn it into a solana wrapped token without knowing how to fmt the address? 
+
+    // neet to tweak contracts on sol to support !ATA?
+    given an xfer from eth=>sol, and without installing sol platform, how do i determine the ATA?
```
nonergodic, line 200: yeah, that one is a tricky bitch. I realized that this would be an issue a while ago but haven't put a lot of thought into it myself yet.

My knee-jerk response is to split platforms into a thin layer that contains light-weight stuff (like address types) and a heavy weight part (if you actually want to do stuff on that chain). But I'm afraid that this is not an abstraction that will actually work/hold up. (If you have to do address translations via RPC lookups like on Sui, you're presumably fucked).

DESIGN.md
```diff
@@ -0,0 +1,205 @@
...
+- A Provider is just an RPC connection and is held by the Chain. Providers are an implementation detail.
+
+## What's a signer vs. a wallet? Should signers have a provider (my answer: no)?
+
+- A Signer is an interface to sign transactions
+- It _may_ be backed by a wallet but not necessarily, as long as it fulfils the interface
+
+## Can we provide some way to make other non-standard applications available to use through the WormholeTransfer?
+
+Say I have an app that defines its own protocol, can I provide something that adheres to the WormholeTransfer interface so a dev can install it and call it like the TokenTransfer?
```
nonergodic, line 182: Good point!

I already thought about this when specifying payload layouts, but until now I didn't realize that essentially layouts for an inheritance hierarchy of sorts... I'll have to think about this more :thinking:

artursapek (approved):

PR #54, opened 2023-09-19
remove noble secp256k1 as dep in signature type def

barnjamin:


evan-gray (approved):

core/definitions/src/signature.ts
```diff
@@ -1,22 +1,12 @@
...
-  constructor(
-    readonly r: bigint,
-    readonly s: bigint,
-    readonly recovery: number
-  ) {
-    super(r, s, recovery);
-  }
+// Signature represents the secp256k1 signature of a Guardian
+export class Signature {
+  constructor(readonly r: bigint, readonly s: bigint, readonly v: number) {}
```
evan-gray, line 3: do you need to store these somewhere?

evan-gray: nvm, now I remember https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

PR #60, opened 2023-09-21
Adding default network from env in connect package

barnjamin:
also making network the primary paramter and allow _overriding_ of conf but default is derived from network

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "clean": "npm run clean --workspaces --if-present",
     "prettier": "npm run prettier --workspaces --if-present",
     "test:integration": "jest --config ./jest.config.ts --roots ./__tests__",
-    "test": "npm run test --workspaces --if-present",
+    "test": "NETWORK=Mainnet npm run test --workspaces --if-present",
     "docs": "typedoc"
   },
   "workspaces": [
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "clean": "npm run clean --workspaces --if-present",
     "prettier": "npm run prettier --workspaces --if-present",
     "test:integration": "jest --config ./jest.config.ts --roots ./__tests__",
-    "test": "npm run test --workspaces --if-present",
+    "test": "NETWORK=Mainnet npm run test --workspaces --if-present",
```
barnjamin, line 32: the nock fixtures were run against mainnet for some reason, will revert this when I update those if this change goes through

platforms/evm/__tests__/integration/tokenBridge.test.ts
```diff
@@ -225,7 +225,8 @@ describe('TokenBridge Tests', () => {
 
       const { transaction } = attestTx;
       expect(transaction.chainId).toEqual(
-        evmNetworkChainToEvmChainId(NETWORK, chain),
+        // @ts-ignore
+        evmNetworkChainToEvmChainId(network, chain),
```
barnjamin, line 229: need to constrain network here to main/testnet because of const mapping

PR #64, opened 2023-09-25
Platform utils

anondev2323:
- removes `parseAddress` from platforms since it is the same functionality as `toNative`
- separates generic (i.e. non-wormhole specific) platform utils into their own module
- adds native decimals and `getCurrentBlock`
- Cleans up `get.../mustGet` contract access methods. Per an earlier discussion, it's preferred to just throw an error

TODO:
- [ ] getBalance should assume the token address is on the correct chain.  Therefore it should accept a token address instead of TokenId/native
- [x] remove parseAddress tests


barnjamin:
> Cleans up get.../mustGet contract access methods. Per an earlier discussion, it's preferred to just throw an error
YEESSSSS

core/definitions/src/testing/mocks/platform.ts
```diff
@@ -40,6 +40,7 @@ export function mockPlatformFactory<P extends PlatformName>(
 export class MockPlatform<P extends PlatformName> implements Platform<P> {
   // @ts-ignore
   readonly platform: P;
+  readonly nativeDecimals = 18n;
```
barnjamin, line 43: can we get this from consts + platform?

It looks like we specify decimals by chain here, maybe it should be platform?
https://github.com/wormhole-foundation/connect-sdk/blob/develop/core/base/src/constants/decimals.ts#L18

platforms/evm/src/platform.ts
```diff
@@ -43,7 +36,18 @@ export module EvmPlatform {
...
+  export const {
+    nativeTokenId,
+    isNativeTokenId,
+    isSupportedChain,
+    getDecimals,
+    getBalance,
+    sendWait,
+    getCurrentBlock,
+    chainFromRpc,
+  } = EvmUtils;
```
anondev2323, line 50: Do you know how to just re-export all of these? Couldn't figure out the right syntax :\

platforms/solana/src/address.ts
```diff
@@ -17,6 +17,9 @@ declare global {
   }
 }
 
+// TODO: is there a zero address for Solana?
+export const SolanaZeroAddress = '11111111111111111111111111111111';
```
anondev2323, line 21: TODO

barnjamin (approved):

PR #66, opened 2023-10-02
Core lint

anondev2323:
Adds lint scripts to core packages.

The primary lint error I address was removing `const` from `const T extends SomeType`. After removing this, I cannot build the packages...


barnjamin:
Closing in favor of #70

PR #73, opened 2023-10-08
idk

barnjamin:


core/base/src/utils/mapping.ts
```diff
@@ -211,12 +211,16 @@ const toMapping = <
     throw new Error("Invalid mapping: empty");
 
   const definedShape = (shape === undefined)
-    ? [range(crr[0].length - 1), [crr[0].length - 1]]
-    : shape.map(ind => typeof ind === "number" ? [ind] : ind);
+    ? [range(crr[0].length - 1), crr[0].length - 1]
```
barnjamin, line 214: removed wrapping by default

core/base/src/utils/mapping.ts
```diff
@@ -211,12 +211,16 @@ const toMapping = <
...
+    : shape
+  const adjustedShape = definedShape.map(ind => typeof ind === "number" ? [ind] : ind);
```
barnjamin, line 216: new var to hold adjusted/normalized shape

core/base/src/utils/mapping.ts
```diff
@@ -211,12 +211,16 @@ const toMapping = <
...
 
+  let finalSingleton = typeof definedShape[1] === 'number';
```
barnjamin, line 218: if the second entry in the requested shape is a number, expect it to be a single value

core/base/src/constants/contracts/cosmos.ts
```diff
@@ -5,13 +5,17 @@ import { Network } from "../networks";
...
-    "wormhole14ejqjyq8um4p3xfqj74yld5waqljf88fz25yxnma0cngspxe3les00fpjx",
+   [
+    ["Wormchain", "wormhole14ejqjyq8um4p3xfqj74yld5waqljf88fz25yxnma0cngspxe3les00fpjx"],
+   ]
   ],
   [
     "Testnet",
-    "wormhole1ctnjk7an90lz5wjfvr3cf6x984a8cjnv8dpmztmlpcq4xteaa2xs9pwmzk",
+    [
+      ["Wormchain", "wormhole1ctnjk7an90lz5wjfvr3cf6x984a8cjnv8dpmztmlpcq4xteaa2xs9pwmzk" ]
```
barnjamin, line 15: adding another layer so it has the same format as other contract configs

PR #82, opened 2023-10-11
Use signed_vaa endpoint on getVaaBytes

mat1asm:
- Uses signed_vaa endpoint, which is also available in wormholescan.
- Adds options to set timeouts.


mat1asm:
replaced by https://github.com/wormhole-foundation/connect-sdk/pull/83

PR #83, opened 2023-10-11
use signed_vaa endpoint on getVaaBytes

mat1asm:
- Uses signed_vaa endpoint, which is also available in wormholescan.
- Adds options to set timeouts.

### DIFF
--- a/connect/src/wormhole.ts
+++ b/connect/src/wormhole.ts
@@ -23,7 +23,7 @@ import {
   WormholeMessageId,
   isTokenId,
 } from "@wormhole-foundation/sdk-definitions";
-import axios, { AxiosResponse } from "axios";
+import axios, { AxiosRequestConfig, AxiosResponse } from "axios";
 
 import { WormholeConfig } from "./types";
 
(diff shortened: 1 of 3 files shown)


connect/src/wormhole.ts
```diff
@@ -391,15 +392,28 @@ export class Wormhole {
     emitter: UniversalAddress | NativeAddress<PlatformName>,
     sequence: bigint,
     retries: number = 5,
+    opts?: {
+      retryDelay?: number;
+      requestTimeout?: number;
+    },
   ): Promise<Uint8Array | undefined> {
```
mat1asm, line 399: Is there a reason not to throw an error containing the status or some error details instead of masking it and just returning undefined? 
In some situations it may be useful for an sdk client to decide depending on that.

barnjamin: No _good_ reason, the retry/timeout stuff here is also begging to be written better, maybe also touching the retry-er thing here: https://github.com/wormhole-foundation/connect-sdk/blob/develop/connect/src/protocols/retry.ts

barnjamin (approved):

PR #104, opened 2023-10-17
add: optimized get token balances

anondev2323:
Right now it just checks for native addresses.  The follow up for this would be to create an `getWrappedAssets` method, so you could call it for wrapped assets as well as native: `getBalances(getWrappedAssets(...))`

### DIFF
--- a/core/definitions/src/testing/mocks/tokenBridge.ts
+++ b/core/definitions/src/testing/mocks/tokenBridge.ts
@@ -1,11 +1,10 @@
 import { PlatformName } from "@wormhole-foundation/sdk-base";
 import {
+  AnyAddress,
   ChainAddress,
   NativeAddress,
   RpcConnection,
   TokenBridge,
-  TokenId,
-  UniversalOrNative,
   UnsignedTransaction,
   VAA,
 } from "../..";
@@ -19,10 +18,10 @@ import {
 export class MockTokenBridge<P extends PlatformName> implements TokenBridge<P> {
   constructor(readonly rpc: RpcConnection<P>) {}
 
-  isWrappedAsset(token: UniversalOrNative<P>): Promise<boolean> {
+  isWrappedAsset(token: AnyAddress): Promise<boolean> {
     throw new Error("Method not implemented.");
   }
-  getOriginalAsset(token: UniversalOrNative<P>): Promise<ChainAddress> {
+  getOriginalAsset(token: AnyAddress): Promise<ChainAddress> {
     throw new Error("Method not implemented.");
   }
   hasWrappedAsset(original: ChainAddress): Promise<boolean> {
@@ -37,7 +36,7 @@ export class MockTokenBridge<P extends PlatformName> implements TokenBridge<P> {
     throw new Error("Method not implemented.");
   }
   createAttestation(
-    address: UniversalOrNative<P>,
+    address: AnyAddress,
   ): AsyncGenerator<UnsignedTransaction> {
     throw new Error("Method not implemented.");
   }
@@ -47,16 +46,16 @@ export class MockTokenBridge<P extends PlatformName> implements TokenBridge<P> {
     throw new Error("Method not implemented.");
   }
   transfer(
-    sender: UniversalOrNative<P>,
+    sender: AnyAddress,
     recipient: ChainAddress,
-    token: "native" | UniversalOrNative<P>,
+    token: "native" | AnyAddress,
     amount: bigint,
     payload?: Uint8Array | undefined,
   ): AsyncGenerator<UnsignedTransaction> {
     throw new Error("Method not implemented.");
   }
   redeem(
-    sender: UniversalOrNative<P>,
+    sender: AnyAddress,
     vaa: VAA<"Transfer"> | VAA<"TransferWithPayload">,
     unwrapNative?: boolean | undefined,
   ): AsyncGenerator<UnsignedTransaction> {
--- a/core/definitions/src/testing/mocks/tokenBridge.ts
+++ b/core/definitions/src/testing/mocks/tokenBridge.ts
@@ -35,9 +35,7 @@ export class MockTokenBridge<P extends PlatformName> implements TokenBridge<P> {
   ): Promise<boolean> {
     throw new Error("Method not implemented.");
   }
-  createAttestation(
-    address: AnyAddress,
-  ): AsyncGenerator<UnsignedTransaction> {
+  createAttestation(address: AnyAddress): AsyncGenerator<UnsignedTransaction> {
     throw new Error("Method not implemented.");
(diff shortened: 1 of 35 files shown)


platforms/cosmwasm/src/address.ts
```diff
@@ -113,7 +114,10 @@ export class CosmwasmAddress implements Address {
   // The denomType is "native", "ibc", or "factory"
   private readonly denomType?: string;
 
-  constructor(address: string | Uint8Array | UniversalAddress) {
+  constructor(address: AnyCosmwasmAddress) {
+    if (address instanceof CosmwasmAddress) {
```
barnjamin, line 118: I'm not positive this `instanceof` check will cause issues but I'm sure its v picky about the check it does. 

If you somehow have a `CosmwasmAddress` from a diff _version_ of the same package it'll fail.  (see https://github.com/algorand/js-algorand-sdk/issues/740 for example)

Mostly pointing it out so if/when we get a weird ticket, this comes to mind easily

anondev2323: What about adding a platform value on the address and check that? `if (address.platform === CosmwasmAddress.platform)`

core/definitions/src/types.ts
```diff
@@ -15,6 +15,16 @@ export type SequenceId = bigint;
 
 export type SignedTx = any;
 
+// TODO: Can we make this more dynamic?
```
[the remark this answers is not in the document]
anondev2323: `AnyAddress<PlatformName>` is kind of what I was thinking of when writing this comment

core/definitions/src/testing/mocks/tokenBridge.ts
```diff
@@ -47,16 +44,16 @@ export class MockTokenBridge<P extends PlatformName> implements TokenBridge<P> {
     throw new Error("Method not implemented.");
   }
   transfer(
-    sender: UniversalOrNative<P>,
+    sender: AnyAddress,
     recipient: ChainAddress,
-    token: "native" | UniversalOrNative<P>,
+    token: "native" | AnyAddress,
```
barnjamin, line 49: AnyAddress has `native`  already, can we get rid of this union type?

core/definitions/src/types.ts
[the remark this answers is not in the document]
barnjamin: ok to drop the todo?

platforms/cosmwasm/src/address.ts
```diff
@@ -102,6 +104,7 @@ function tryDecode(data: string): { data: Uint8Array; prefix?: string } {
 export class CosmwasmAddress implements Address {
   static readonly contractAddressByteSize = 32;
   static readonly accountAddressByteSize = 20;
+  public readonly platform: PlatformName = 'Cosmwasm';
```
barnjamin, line 107: possible to take this from `CosmwasmPlatform.platform`?

platforms/cosmwasm/src/address.ts
```diff
@@ -224,6 +235,11 @@ export class CosmwasmAddress implements Address {
     return true;
   }
 
+  static instanceof(address: any) {
+    const platform: PlatformName = 'Cosmwasm';
+    return address.platform === platform;
```
barnjamin, line 240: same deal, can we take this str from the platform's static const?

platforms/cosmwasm/src/platformUtils.ts
```diff
@@ -79,10 +85,31 @@ export module CosmwasmUtils {
...
 
-    const { amount } = await rpc.getBalance(walletAddress, token.toString());
+    const addrStr = new CosmwasmAddress(token).toString();
+    const { amount } = await rpc.getBalance(walletAddress, addrStr);
     return BigInt(amount);
   }
 
+  export async function getBalances(
+    chain: ChainName,
+    rpc: QueryClient & BankExtension & IbcExtension,
```
barnjamin, line 95: Should we keep this one as the default `CosmWasmClient` and convert to the rpc type we need in the body of the fn?

platforms/solana/src/chain.ts
```diff
@@ -3,27 +3,27 @@ import {
...
     address: UniversalAddress,
   ): Promise<NativeAddress<'Solana'>> {
     const tb = await this.getTokenBridge();
 
-    const mintAddress: UniversalOrNative<'Solana'> =
+    const mintAddress =
       token === 'native'
         ? await tb.getWrappedNative()
-        : token.toUniversalAddress();
+        : new SolanaAddress(token);
```
barnjamin, line 26: This gets called from the wormhole class to prep a transfer, I'm not positive but it _may_ cause issues if the address passed has not been "universal-ified"

barnjamin (commented):
Awesome! looks great, just a few nits

platforms/cosmwasm/src/constants.ts
```diff
@@ -70,7 +70,7 @@ export const cosmwasmNetworkChainToChainId = constMap(
   networkChainCosmwasmChainIds,
 );
 
-const cosmwasmAddressPrefix = [
+export const cosmwasmAddressPrefix = [
```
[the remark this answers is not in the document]
anondev2323: I actually don't need this anymore.  I realized that `fromBech32` from the `@cosmjs/encoding` handles the prefix stuff

core/definitions/src/testing/mocks/tokenBridge.ts
anondev2323: string is already supported, so I just removed 'native'

platforms/solana/src/chain.ts
anondev2323: Ah good point, I'll revert for now.  Maybe in the future, we should just throw an error if they try to send to Solana but don't have the Solana package installed.  Just add some more safety guardrails

platforms/solana/src/address.ts
```diff
@@ -8,6 +8,7 @@ import {
 
 import { PublicKey } from '@solana/web3.js';
 import { AnySolanaAddress } from './types';
+import { SolanaPlatform } from '../dist/esm';
```
barnjamin, line 11: This will probably break stuff in build

platforms/solana/src/chain.ts
barnjamin: good idea re error on no Solana, hoping this specific case can be eliminated once the new Solana contracts go to Mainnet since it allows sending to standard acct not ATA

platforms/solana/src/address.ts
anondev2323: Absolutely, I must have hit tab on the wrong line cause I did not mean to import from dist :|

barnjamin (approved):

PR #109, opened 2023-10-19
finalize automatic payload discrimination and rework VAA handling

nonergodic:
also added/changed:
* renamed payloadliterals to include the protocol name (to get rid of the overly general Transfer payload and instead have it now as "TokenBridge-Transfer", etc.)
* added handling for such protocol-payloadName composite payload literals
* added CCTP governance payloads
* implemented lazyInstantiation helper (currently has a rather questionable placing in misc.ts)
* introduced a TokenBridge namespace for associated types/functions

still oustanding:
* not all free standing functions have been absorbed into the TokenBridge namespace
* CoreBridge does not use a namespace like TokenBridge yet
* there's more cleanup to be done (still a bunch of redundant code left)

### DIFF
--- a/__tests__/tilt/tokenbridge.test.ts
+++ b/__tests__/tilt/tokenbridge.test.ts
@@ -65,7 +65,7 @@ describe("Tilt Token Bridge Tests", () => {
             const [msgid] = await eth.parseTransaction(txns[txns.length - 1].txid)
             expect(isWormholeMessageId(msgid)).toBeTruthy()
 
-            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "AttestMeta")
+            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "TokenBridge-AttestMeta")
             expect(vaa).toBeTruthy()
 
             const completeAttest = stb.submitAttestation(vaa!, solAcct.address as SolanaAddress)
--- a/__tests__/tilt/tokenbridge.test.ts
+++ b/__tests__/tilt/tokenbridge.test.ts
@@ -65,7 +65,7 @@ describe("Tilt Token Bridge Tests", () => {
             const [msgid] = await eth.parseTransaction(txns[txns.length - 1].txid)
             expect(isWormholeMessageId(msgid)).toBeTruthy()
 
-            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "TokenBridge-AttestMeta")
+            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "TokenBridge:AttestMeta")
             expect(vaa).toBeTruthy()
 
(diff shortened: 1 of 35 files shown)


__tests__/tilt/tokenbridge.test.ts
```diff
@@ -65,7 +65,7 @@ describe("Tilt Token Bridge Tests", () => {
             const [msgid] = await eth.parseTransaction(txns[txns.length - 1].txid)
             expect(isWormholeMessageId(msgid)).toBeTruthy()
 
-            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "AttestMeta")
+            const vaa = await wh.getVAA(msgid.chain, msgid.emitter, msgid.sequence, "TokenBridge-AttestMeta")
```
barnjamin, line 68: Can we use a diff char to delimit namespace-payload type? maybe `:` or `/` or `.`?  

I don't have a good reason besides a hyphen feeling wrong 😆

core/definitions/src/vaa.ts
```diff
@@ -90,6 +117,28 @@ export interface VAA<PL extends PayloadLiteral = PayloadLiteral>
...
 
+//We enforce distribution over union types, e.g. have
+//    ProtocolVAA<"TokenBridge", "Transfer" | "TransferWithPayload">
+//  turned into
+//    VAA<"TokenBridge-Transfer"> | VAA<"TokenBridge-TransferWithPayload">
+//  rather than
+//    VAA<"TokenBridge-Transfer" | "TokenBridge-TransferWithPayload">
+//  because while the latter is considered more idiomatic/canonical, it actually interferes with
+//  the most natural way to narrow VAAs via querying the payloadName or payloadLiteral.
+//  (Thanks for absolutely nothing, Typescript).
```
barnjamin, line 128: 🥇

core/definitions/src/vaa.ts
```diff
@@ -90,6 +117,28 @@ export interface VAA<PL extends PayloadLiteral = PayloadLiteral>
...
+//  For example, given the TokenBridge VAA union example:
+//  if (vaa.payloadName === "Transfer")
+//    typeof vaa //no narrowing - still resolves to the union type when using the latter approach
```
barnjamin, line 131: can we do a type guard fn for this that works?

nonergodic: Yeah, easily, though it's still natural (most of the time you'll probably want to narrow based on payloadName, but maybe sometimes also by payloadLiteral or maybe even protocolName...?)

Do you have anything in particular in mind why you would want a dedicated type guard?

barnjamin: Not really, just thinking about how to get proper narrowing


Maybe useful someplace like this https://github.com/wormhole-foundation/connect-sdk/blob/develop/platforms/evm/src/protocols/tokenBridge.ts#L244-L251 

That would become `if(TokenBridge.isTransferWithPayload(vaa))...` but def not critical. 

We can always add it later

PR #147, opened 2023-11-09
add address format conversion for constant addresses

nonergodic:


core/base/src/utils/encoding.ts
```diff
@@ -1,5 +1,7 @@
 import { base16, base64, base58 } from '@scure/base';
 
+export { bech32 } from '@scure/base';
```
barnjamin, line 3: I expect this will work fine but can we just export it the same way as others?

core/base/src/utils/encoding.ts
```diff
@@ -41,3 +43,17 @@ export const toUint8Array = (value: string | bigint): Uint8Array =>
...
+export const concat = (...args: Uint8Array[]): Uint8Array => {
+  const length = args.reduce((acc, curr) => acc + curr.length, 0);
+  const result = new Uint8Array(length);
+  let offset = 0;
+  args.forEach((arg) => {
+    result.set(arg, offset);
+    offset += arg.length;
+  });
+  return result;
+}
```
barnjamin, line 59: should we namespace the u8array stuff as well?

core/base/src/utils/misc.ts
```diff
@@ -6,3 +6,22 @@ export function lazyInstantiate<T>(factory: () => T): () => T {
     return instance;
   }
 }
+
+export function onlyOnce<T extends []>(fn: (...args: T) => any, ...args: T): () => void {
```
barnjamin, line 10: intedesting

core/definitions/src/universalAddress.ts
```diff
@@ -40,20 +38,39 @@ export class UniversalAddress implements Address {
...
+    return encoding.equals(this.address, other.address);
   }
 
-  static isValidAddress(address: string) {
-    return (
-      encoding.hex.valid(address) &&
-      encoding.stripPrefix("0x", address).length === UniversalAddress.byteSize * 2
-    );
+  static isValidAddress(address: string, format: PlatformAddressFormat = "hex") {
+    return !throws(() => UniversalAddress.stringToUint8Array(address, format));
```
barnjamin, line 45: pretty

barnjamin (approved):

core/base/src/utils/encoding.ts
nonergodic: The issue is that bech32 is rather different to the rest because it's not a simple one to one conversion. It takes the human readable descriptor, the actual data, and then it computes a payload (and vice versa).

I'm already using the slightly awkward `decodeWithBytes` method (or whatever its exact name was).

I'd have made it uniform if I had seen a straigh-forward way how.

core/base/src/utils/encoding.ts
nonergodic: I think so. I don't think we want to have a free-standing `equals` and `concat` lying around.

core/base/src/utils/misc.ts
nonergodic: This is for all the factory registration try - catch stuff. I still have it only locally. I'd really like to figure out why the same module code is ever executed twice. This sounds like the real issue to me.

PR #151, opened 2023-11-11
DRAFT - Algorand Platform & TokenBridge

SilentRhetoric:
This is a first draft that adds Algorand as a platform and a sub-package for TokenBridge.


platforms/algorand/protocols/tokenBridge/src/constants.ts
barnjamin, line 1: wat

Issue #114, opened 2023-10-22
Implement Algorand platform


barnjamin:
Implement the Algorand platform and TokenBridge protocol

SilentRhetoric:
Planning to take a look at this

seichris:
Looking into it already

barnjamin:
@SilentRhetoric @seichris please consider collaboration if both interested, ultimately only one implementation can be merged 😬 

Please keep this issue updated with progress

VKappaKV:
Hello everyone, looking into this, made any progresses already?

SilentRhetoric:
A PR #151 has been created with a first draft of this implementation.  No tests/examples, yet.

SilentRhetoric:
@cNikolaou @VKappaKV @seichris I invite you to reach out to me via the email address on my GitHub profile so we can discuss potential collaboration.

PR #152, opened 2023-11-12
Add Algorand Platform

cNikolaou:
The current PR contains an implementation of the Algorand Platform and the `TokenBridge` protocol, which depends on functionality moved from: 
- https://github.com/wormhole-foundation/wormhole/tree/38f02fa2fc8e80fc3af213959fe397cc6a936778/sdk/js/src/algorand
- https://github.com/wormhole-foundation/wormhole/tree/38f02fa2fc8e80fc3af213959fe397cc6a936778/sdk/js/src/token_bridge

after making the necessary changes to fit the current software architecture.

It doesn’t currently include tests and I am currently experimenting with it to fix issues.

However, I would like to know:

1. Algorand’s tokens are represented by integers so they didn’t exactly fit under the current `UniversalAddress` interface. Are you thinking of including the `bigint` in the union for `UniversalAddress` or should the `assetId` and `appId` addresses be passed in a different format? Currently I change the types of some of the function arguments in `core/definitions/src/protocols/tokenBridge.ts`.

2. Some of the functionality regarding the VAAs sends transactions that seem to be for a different use case (eg “setting up a new guardian set” - you can find a comment like that in the code moved from the `@wormhole-foundation/whormhole/sdk/js/src/algorand.ts`). I am looking whether there are transactions that can be removed.

3. I tried to format the transactions that need to be sent based on the Solana/Evm interface implementation that relies on generators. I was thinking of using the `algosdk.AtomicTransactionComposer` to group some transactions (eg the ones in `submitVAAHeader` or the `optin`) together and send an atomic transaction. Is that a good approach or it is better to follow the current model of sending each transaction at once from a generator?

4. Related to that is how to handle the transaction fees. There are some computations of the fees but I am not sure whether they depend on a different sing and send model. 

5. Currently the implementation of `getWrappedAsset` for the `AlgorandTokenBridge`. I am looking into Wormhole's smart contract to find out where the address of the token is stored but I think I need to do some byte manipulation to do that.

6. Is it of interest to copy the necessary files from `wormhole-foundation/wormhole/sdk` under the `platforms/algorand/tokenBridge/utils/` to preserver the history instead of copying and modifying the code directly?


barnjamin:
Hello! Answering the questions below BUT there is a PR up for this same Algorand implementation bounty here https://github.com/wormhole-foundation/connect-sdk/pull/151.

Ultimately I can only merge 1 implementation so please consider collaboration with @SilentRhetoric


> Algorand’s tokens are represented by integers so they didn’t exactly fit under the current UniversalAddress interface. Are you thinking of including the bigint in the union for UniversalAddress or should the assetId and appId addresses be passed in a different format? Currently I change the types of some of the function arguments in core/definitions/src/protocols/tokenBridge.ts.

Let's not change the interface in core but we can make it more generic in the Algorand implementation for TokenBridge

> Some of the functionality regarding the VAAs sends transactions that seem to be for a different use case (eg “setting up a new guardian set” - you can find a comment like that in the code moved from the @wormhole-foundation/whormhole/sdk/js/src/algorand.ts). I am looking whether there are transactions that can be removed.

These _should_ be supported ultimately, this sdk should subsume all the functionality of the current wormhole sdk.

> I tried to format the transactions that need to be sent based on the Solana/Evm interface implementation that relies on generators. I was thinking of using the algosdk.AtomicTransactionComposer to group some transactions (eg the ones in submitVAAHeader or the optin) together and send an atomic transaction. Is that a good approach or it is better to follow the current model of sending each transaction at once from a generator?

You should be able to use the asyncgenerator pattern in the same way as others. The ATC should come into play inside the AlgorandSigner implementation. That is, in `sign(UnsignedTx[])` it should pack the ATC and sign the transactions then return the signed txns.

> Related to that is how to handle the transaction fees. There are some computations of the fees but I am not sure whether they depend on a different sing and send model.

This (like eth with gas fees) is up to the implementation of the `Signer`

> Currently the implementation of getWrappedAsset for the AlgorandTokenBridge. I am looking into Wormhole's smart contract to find out where the address of the token is stored but I think I need to do some byte manipulation to do that.

I'm not sure offhand but it should be somewhere in the wormhole monorepo

> Is it of interest to copy the necessary files from wormhole-foundation/wormhole/sdk under the platforms/algorand/tokenBridge/utils/ to preserver the history instead of copying and modifying the code directly?

No, we can look back at git history on monorepo if necessary

barnjamin:
For viz: #114

PR #158, opened 2023-11-13
Implement Algorand Platform and Protocols

cNikolaou:
This succeeds the PR #152 and parts from #151 will be merged here.

Things that will need to change:
- The `address: string` to `address: Uint8` on `AlgorandAddress`, alongside all the functions to get the string format of the address.
- Revert the type of `nativeAddress` in `core/definitions/src/protocols/tokenBridge.ts` to `AnyAddress` from `AnyAddress | bigint`.

What needs to be implemented:
- Find a way to handle the ASAs which do not have `addresses: string` but `id: bigint`
- `getWrappedAsset()` on `platforms/algorand/protocols/tokenBridge/src/tokenBridge.ts`


cNikolaou:
Currently, this PR implements:
- `TokenTransfer` from Algorand to the other supported chains and vice versa. This allows native tokens and other tokens to be transferred from and to Algorand. However, for that to work the metadata of the token that is going to be transfered need to be attested on both chains so that the `destinationChain` is aware of the `decimals`, `symbol`, `name` of the token.
- `AttestationTransfer` from any chain to any other chain, including Algorand. This allows `AttestMeta` VAAs to be attested (`createAttestation()`) on the `sourceChain` and submitted (`submitAttestation()`) on the `destinationChain`. This requires further work since some chain pairs currently fail.

The failure in transferring attestations between some chain pairs is most likely related to the format of the data passed between the chains. For example, the `submitAttestation()` fails when submitting an `AttestMeta` VAA from Avalanche to Solana, but not from Avalanche to Algorand. Further investigation and a more comprehensive analysis is required.

PR #178, opened 2023-12-20
Refactor Addresses

barnjamin:
A constant source of confusion is the `Address`/`UniversalAddress`/`NativeAddress<C>` types.

This PR removes the option of passing a `Platform` as the generic type parameter to help with debugging.

It also removes the `nativeChainAddress` method in favor of the static Wormhole class method to parse addresses or create a `ChainAddress`

### DIFF
--- a/core/definitions/src/protocols/ibc.ts
+++ b/core/definitions/src/protocols/ibc.ts
@@ -1,13 +1,12 @@
 import {
   Chain,
   ChainId,
-  ChainToPlatform,
   Network,
   Platform,
   PlatformToChains,
   encoding,
   toChain,
-  toChainId
+  toChainId,
 } from "@wormhole-foundation/sdk-base";
 import { AccountAddress, ChainAddress, NativeAddress, TokenAddress } from "../address";
 import { IbcMessageId, WormholeMessageId } from "../attestation";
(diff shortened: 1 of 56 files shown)


artursapek (commented):
I hate to be that guy, but every example [here](https://www.typescriptlang.org/docs/handbook/namespaces.html) has capitalized namespaces 😬

core/definitions/src/protocols/ibc.ts
```diff
@@ -153,8 +135,8 @@ export function makeGatewayTransferMsg<CN extends Chain>(
   const address =
     typeof recipient === "string"
       ? recipient
-      // @ts-ignore
-      : encoding.b64.encode(recipient.toString());
+      : // @ts-ignore
+        encoding.b64.encode(recipient.toString());
```
barnjamin, line 139: This one is still whack, `recipient` is `never` type 🤔

nonergodic: This makes sense in the context of the definition layer, because at this point, no native address has been registered yet.

Using declaration merging with a global namespace and hence creating a dynamic type of sorts is not something that typescript "really thought of", I believe. It's quite the abuse of the system akin to early stage C++ template metaprogramming, when nobody had yet realized that they were Turing complete until the first guy with a "I bet I can make this baby run doom" mindset came around.

PR #195, opened 2024-01-03
Solana: Provide ability to resign transactions with ephemeral keys

barnjamin:
Solana transaction changes to make it possible to resubmit a tx

This requires us to pass not only the transaction but all the ephemeral keypairs used to sign the transactions so that the transaction may be updated with a more recent blockhash and re-signed and sent.


0xMMorty:
#752 When I used the latest version code to implement the transaction, it failed. It seems that this piece of code might work here, but the new version code doesn't seem to have this logic anymore, so I failed to get the block height and such. Can you confirm?

PR #197, opened 2024-01-03
Use union types for TransferReceipt

artursapek:
This change makes the TransferReceipt code a little more verbose but a lot more legible. And the compile-time checks seems stricter now, because `TransferState` is only kept in one place. So that's nice.

### DIFF
--- a/connect/src/protocols/cctpTransfer.ts
+++ b/connect/src/protocols/cctpTransfer.ts
@@ -31,10 +31,6 @@ import { signSendWait } from "../common";
 import { DEFAULT_TASK_TIMEOUT } from "../config";
 import { Wormhole } from "../wormhole";
 import {
-  AttestedTransferReceipt,
-  CompletedTransferReceipt,
-  SourceInitiatedTransferReceipt,
-  SourceFinalizedTransferReceipt,
   TransferQuote,
   TransferReceipt,
   TransferState,
@@ -42,6 +38,7 @@ import {
   isSourceInitiated,
   isSourceFinalized,
   isAttested,
+  AttestedTransferReceipt,
 } from "../wormholeTransfer";
 
 type CircleTransferProtocol = "CircleBridge" | "AutomaticCircleBridge";
@@ -489,9 +486,7 @@ export class CircleTransfer<N extends Network = Network>
 
     const originTxs = xfer.txids.filter((txid) => txid.chain === xfer.transfer.from.chain);
     if (originTxs.length > 0) {
-      receipt = { ...receipt, state: TransferState.SourceInitiated, originTxs } as Partial<
-        SourceInitiatedTransferReceipt<CircleTransferProtocol, Chain, Chain>
-      >;
+      receipt = { ...receipt, state: TransferState.SourceInitiated, originTxs };
     }
 
     const att = xfer.attestations?.filter((a) => isWormholeMessageId(a.id)) ?? [];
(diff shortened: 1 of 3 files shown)


connect/src/wormholeTransfer.ts
```diff
@@ -41,64 +41,79 @@ export enum TransferState {
 }
 
 // Base type for common properties
-interface BaseTransferReceipt<PN extends ProtocolName, SC extends Chain, DC extends Chain> {
+export interface BaseTransferReceipt<PN extends ProtocolName, SC extends Chain, DC extends Chain> {
```
barnjamin, line 44: does this need to be exported?

connect/src/wormholeTransfer.ts
```diff
@@ -41,64 +41,79 @@ export enum TransferState {
...
-  ? CompletedTransferReceipt<PN, SC, DC>
-  : TS extends TransferState.Attested
-  ? AttestedTransferReceipt<PN, SC, DC>
-  : TS extends TransferState.SourceFinalized
-  ? SourceFinalizedTransferReceipt<PN, SC, DC>
-  : TS extends TransferState.SourceInitiated
-  ? SourceInitiatedTransferReceipt<PN, SC, DC>
-  : never;
+> =
+  | BaseTransferReceipt<PN, SC, DC>
```
barnjamin, line 112: oh maybe needs to be exported because of this?

connect/src/wormholeTransfer.ts
artursapek: yes because otherwise it complains that `track` leaks a private type

```
error TS4050: Return type of public static method from exported class has or is using name 'BaseTransferReceipt' from external module "/Users/artur/work/connect-sdk/connect/src/wormholeTransfer" but cannot be named.
```

connect/src/wormholeTransfer.ts
artursapek: oh right. I could try to get rid of this

connect/src/protocols/cctpTransfer.ts
```diff
@@ -487,7 +493,9 @@ export class CircleTransfer<N extends Network = Network>
 
     const originTxs = xfer.txids.filter((txid) => txid.chain === xfer.transfer.from.chain);
     if (originTxs.length > 0) {
-      receipt = { ...receipt, state: TransferState.SourceInitiated, originTxs };
+      receipt = { ...receipt, state: TransferState.SourceInitiated, originTxs } as Partial<
+        SourceInitiatedTransferReceipt<CircleTransferProtocol, Chain, Chain>
```
barnjamin, line 497:
```suggestion
        SourceInitiatedTransferReceipt<CircleTransferProtocol, typeof from.chain, typeof to.chain>
```

connect/src/protocols/cctpTransfer.ts
```diff
@@ -33,15 +33,15 @@ import { Wormhole } from "../wormhole";
...
+  SourceInitiatedTransferReceipt,
   TransferQuote,
   TransferReceipt,
   TransferState,
   WormholeTransfer,
-  isSourceInitiated,
-  isSourceFinalized,
   isAttested,
+  isSourceFinalized,
+  isSourceInitiated,
```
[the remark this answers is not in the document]
barnjamin: unsure tbh, might be an editor thing

PR #209, opened 2024-01-10
fix signature packing for algorand

barnjamin:


artursapek (commented):
I dont understand this change. are there tests that should have caught this? if not, we should add some.

barnjamin:
@artursapek this code happens to work when there is only 1 sig (testnet/devnet) but will fail if the loop runs multiple times since it'll quickly get out of the valid signature range

artursapek (approved):

PR #212, opened 2024-01-11
Add sui platform

barnjamin:


platforms/sui/eslintrc.json
scnale, line 1: Shouldn't this file be named `.eslintrc.json`?

barnjamin: yes, looks like this was copy/pasta'd with no checking to see if actually worked

Set it up (mostly) correctly here but not a prereq to commit yet
https://github.com/wormhole-foundation/connect-sdk/pull/331

PR #217, opened 2024-01-12
routes: add token bridge route

barnjamin:


artursapek:
I would just store `options` in the class instance at this point. pass it into constructor

connect/src/routes/tokenBridgeRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export type TokenBridgeTransferOptions = {
+  payload?: Uint8Array;
+};
+
+export class TokenBridgeRoute<N extends Network> extends Route<N, TokenBridgeTransferOptions> {
+  NATIVE_GAS_DROPOFF_SUPPORTED = false;
+  IS_AUTOMATIC = false;
+
+  constructor(wh: Wormhole<N>, request: TransferRequest) {
```
artursapek, line 20: I think you could just not have this function here, and `super` will take it automatically

connect/src/routes/tokenBridgeRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // If the destination token was set, and its different than what
+    // we'd get from a token bridge transfer, then this route is not supported
+    if (this.request.destination) {
+      const destToken = await TokenTransfer.lookupDestinationToken(
+        fromChain,
+        toChain,
+        this.request.source,
+      );
+      if (!isSameToken(destToken, this.request.destination)) return false;
```
artursapek, line 53: Connect's code has all kinds of special cases I don't understand. I think we need it in this SDK too.

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/bridge/bridge.ts#L48-L62

PR #218, opened 2024-01-12
add automatic token bridge, modify route types

barnjamin:


connect/src/routes/automaticTokenBridgeRoute.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  async validate(
+    options: AutomaticTokenBridgeRoute.Options,
+  ): Promise<ValidationResult<boolean, ValidationError>> {
+    const transfer = this.toTransferDetails(options);
+    try {
+      await TokenTransfer.validateTransferDetails(this.wh, transfer);
+      return { valid: true };
+    } catch (e) {
+      return { valid: false, error: "Some other error" };
```
artursapek, line 64: maybe `ValidationResult` should allow `error` to be `any`?

barnjamin: I just made it `Error` since we have it already there

PR #219, opened 2024-01-13
Routes: Standardize interface

barnjamin:
Move around where we do different parts of the validation.

Current thinking:

`isSupported` should be fast and check only whether or not the route is supported, not async, static definition lets us skip init if we can filter off it (implemented)

`isAvailable` async, check the chain (or cache) eg for token in some approved list

`validate` async, make sure opts + request are valid, check fees and token destination existing

`quote` async, get a quote for dst token and amount, fee, dest native gas, slippage?

`getDefaultOptions` undecided, should this be async and fetch fee and then make fee part of the options?


connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -11,72 +12,104 @@ import { AutomaticRoute, ValidationResult } from "../route";
...
+      const quote = await this.quote(options);
+      const { amount: fee } = quote.relayFee!;
+
+      if (amount < fee) throw new Error(`Amount must be greater than fee:  ${amount} > ${fee}`);
+
+      options.nativeGas = options.nativeGas ?? 0n;
+
+      if (destination) {
+        if (isTokenId(destination)) {
+          if (!isSameToken(quote.destinationToken.token, destination))
```
artursapek, line 64: I don't understand why you moved this out of `isSupported`. what is the purpose of `isSupported` now? and when is `validate` supposed to be called?

in Connect, `isSupported` and `isAvailable` are what is used to filter the routes so that the UI can show all of the available routes. a basic check like this is required to show whether TokenBridge is a supported route that the user can choose. pushing this into validate might make `isSupported` fast, but it also makes it pretty useless.

my conceptual understanding is this:

`isSupported`: can this route theoretically perform this transfer? this involves the selected chains, tokens, and amount.
`isAvailable`: check in with any online dependencies, like a relayer, to make sure they're repsonsive.
`validate`: once the user selects a supported & available route, they put in options. this checks if they are correct. the amount is not an option, it's a required input. same with the tokens.

connect/src/routes/route.ts
```diff
@@ -1,45 +1,78 @@
...
+    wh: Wormhole<N>,
+    request: RouteTransferRequest,
+    fromChain?: ChainContext<N>,
+    toChain?: ChainContext<N>,
+  ): Route<N, OP>;
+  // Check if this route is supported for the given transfer request
+  // e.g. check if the protocols on the specific chains are supported
+  isSupported(fromChain: ChainContext<N, Platform>, toChain: ChainContext<N, Platform>): boolean;
+  // Get the default options for this route, useful to prepopulate a form
+  getDefaultOptions(): OP;
```
artursapek, line 35: I think having this be a static method makes sense, but I disagree with what you did to `isSupported`

connect/src/routes/tokenBridge/automatic.ts
barnjamin: I'm thinking of `isSupported` as just a check to see if this route can even be created. If the protocols or chains passed are not supported, filter it out immediately.

I'm thinking of `isAvailable` as a filter on the input tokens. 

I'm thinking of `validate` as the last step before sending, modifying and passing back suggested options.


It could be that the methods aren't broken out perfectly but I think it does make sense to have at least 1 very fast check to filter out as many potential routes as possible before doing extra work like reaching out to ask the contract for details.

artursapek: yeah I'm open to any interface here, I literally copied this out of Connect. it's probably not ideal. all I know is, the UI has to not show routes that can't handle your inputs. in this case, you would show TokenBridge as a route before checking if the tokens are the same, because that check is inside `validate`. so that doesn't work.

> Maybe we do need something like isRouteSupported and isTokenPairSupported where the latter does the work of checking to see if its a valid path from src->dst tokens?

based on my read of Connect's code, we generally need to make sure of these things before showing a route as a choice in the UI:

- source and destination chains are supported
- source and destination tokens are supported

barnjamin: >source and destination chains are supported

this can be `isSupported`

> source and destination tokens are supported

source token support can be `isAvailable` by checking acceptable input tokens

if the Destination token is set, some awkward logic associated:

If the user passes `native`, despite not having the ability to "swap" to anything, the AutoTB can do native gas dropoff. The `nativeGas` param in Options _must_ be set to one of `0n`/`undefined`/`amount - fee` to be sure no "extra" (non gas tokens) tokens are delivered. 

If the user passes anything else as a destination token, we can lookup the expected destination token and compare it. 

Both the fee and destination token are returned as part of fetching a quote.

> amount is not too small or too large

We can filter on truncated(amount) == 0 but we'll have to fetch a quote to determine if the fee is > amount. 

We may want to grab a quote early on, in `isAvailable` or some new async static initializer for example. 

The quote may be cache-able but should be considered invalid if the options are different than the last time we asked for a quote, in that case it needs to be re-fetched.

artursapek: I still don't understand the purpose of `isSupported` if all it does is check the two chains. you'll never just call that and be done; you always need to check more stuff before you can return a set of routes from the resolver.

connect/src/routes/route.ts
```diff
@@ -12,25 +12,16 @@ import { TransferQuote, TransferReceipt } from "../wormholeTransfer";
...
   from: ChainAddress;
   to: ChainAddress;
-  amount: bigint;
   source: TokenId | "native";
   destination?: TokenId | "native";
 }
 
 export type ValidationResult<T> =
-  | { quote: TransferQuote; options: T; valid: true }
+  | { quote: TransferQuote; amount: bigint; options: T; valid: true }
```
barnjamin, line 20: y here and not in the options type?

connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -57,62 +59,52 @@ export class AutomaticTokenBridgeRoute<N extends Network> extends AutomaticRoute
...
         //  if it came in set
         //  and they specified native dest token
         //  then nativeGas must be exactly the same as amount - fee
-        if (options.nativeGas !== amount - fee)
-          throw new Error("Native gas cannot be greater than amount - fee");
       }
 
-      if (options.nativeGas < 0n) throw new Error("Native gas cannot be negative");
+      if (nativeGas > 1.0 || nativeGas < 0.0) {
+        throw new Error("Native gas must be between 0% and 100%");
```
barnjamin, line 81: potentially confusing error message if they pas "50"

connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -133,18 +125,46 @@ export class AutomaticTokenBridgeRoute<N extends Network> extends AutomaticRoute
...
     const transfer = {
       from,
       to,
       amount,
       token: source,
       automatic: true,
-      ...options,
+      options: {
+        nativeGas,
+      },
```
barnjamin, line 167: this wont work since `nativeGas` shouldn't be nested in options for `TokenTransferRequest`

connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -133,18 +125,46 @@ export class AutomaticTokenBridgeRoute<N extends Network> extends AutomaticRoute
...
     const { source, from, to } = this.request;
+    const amount = await this.stringToBigInt(params.amount);
+    let options = params.options ?? AutomaticTokenBridgeRoute.getDefaultOptions();
+
+    // Determine nativeGas
+    let nativeGas = 0n;
+
+    // Calculate nativeGas in base units if options.nativeGas isn't 0
+    if (options && options.nativeGas > 0) {
+      const quote = await this.quote(params);
```
barnjamin, line 147: does this do an infinite recursion since we call toTransferDetails in `quote`?

connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -133,18 +125,46 @@ export class AutomaticTokenBridgeRoute<N extends Network> extends AutomaticRoute
     yield* TokenTransfer.track(this.wh, receipt, timeout, this.fromChain, this.toChain);
   }
 
-  private toTransferDetails(
-    amount: bigint,
-    options: AutomaticTokenBridgeRoute.Options,
-  ): TokenTransferDetails {
+  private async stringToBigInt(amount: string = "0"): Promise<bigint> {
```
barnjamin, line 128: Instead of doing this here, lets grab the token details (name, decimals, ...) right before searching for routes and then cache that in the Route obj 

Then we don't have to make converter async

connect/src/routes/tokenBridge/automatic.ts
barnjamin: instead of getting a full `quote` we can ask for _just_ the fee from the auto tb?

artursapek: lmfao oops

connect/src/routes/tokenBridge/automatic.ts
artursapek: ah my b

PR #224, opened 2024-01-16
Core: move ProtocolName

barnjamin:
... from base to definitions, provide interface to register


barnjamin:
moved to https://github.com/wormhole-foundation/connect-sdk/pull/225

PR #228, opened 2024-01-17
WIP: CCTP routes

artursapek:
pls no bully


connect/src/routes/cctp/manual.ts
```diff
@@ -0,0 +1,130 @@
...
+type Vr = ValidationResult<Op>;
+
+export class CCTPRoute<N extends Network> extends ManualRoute<N, Op> {
+  static getDefaultOptions(): Op {
+    return {
+      payload: undefined,
+    };
+  }
+
+  async isSupported(): Promise<boolean> {
```
barnjamin, line 34: This should be checking if the chain context returns true for `supportsCircleBridge`

PR #236, opened 2024-01-19
Routes: define interface for generic params

barnjamin:
Improve typing on generic route params

### DIFF
--- a/connect/src/routes/route.ts
+++ b/connect/src/routes/route.ts
@@ -2,31 +2,33 @@ import { Network } from "@wormhole-foundation/sdk-base";
 import { Signer, TransactionId } from "@wormhole-foundation/sdk-definitions";
 import { Wormhole } from "../wormhole";
 import { RouteTransferRequest } from "./request";
+import {
+  Options,
+  Quote,
+  Receipt,
+  TransferParams,
+  ValidatedTransferParams,
+  ValidationResult,
+} from "./types";
 
-export interface TransferParams<OP> {
-  amount: string;
-  options?: OP;
-}
-
-export interface ValidatedTransferParams<OP> extends Required<TransferParams<OP>> {}
-
-export type ValidationResult<OP> =
-  | { params: ValidatedTransferParams<OP>; valid: true }
-  | { params: TransferParams<OP>; valid: false; error: Error };
-
-export type RouteConstructor<N extends Network, OP> = {
+export type UnknownRouteConstructor<N extends Network> = RouteConstructor<N>;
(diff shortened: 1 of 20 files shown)


connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -43,7 +43,7 @@ type R = TransferReceipt<"AutomaticTokenBridge">;
 export class AutomaticTokenBridgeRoute<N extends Network> extends AutomaticRoute<N, Op, R, Q> {
   NATIVE_GAS_DROPOFF_SUPPORTED = true;
 
-  static getDefaultOptions(): Op {
+  getDefaultOptions(): Op {
```
barnjamin, line 46: removed static so the same Op param we pass in generics is used

connect/src/routes/route.ts
```diff
@@ -14,19 +16,54 @@ export type ValidationResult<OP> =
...
+    token: TokenId;
+    amount: bigint;
+  };
+  relayerFee?: {
+    token: TokenId;
+    amount: bigint;
+  };
+  // swap slippage? what else?
+  // native gas dropoff?
+  // estimate of _time_?
```
barnjamin, line 46: 👀

connect/src/routes/route.ts
```diff
@@ -14,19 +16,54 @@ export type ValidationResult<OP> =
...
+}
+
+export interface Quote {
+  sourceToken: {
+    token: TokenId;
+    amount: bigint;
+  };
+  destinationToken: {
+    token: TokenId;
+    amount: bigint;
```
artursapek, line 38: do we need to complicate this any further for swaps? swaps usually will give a minimum amount (worst case), and an estimated amount. I suppose they could just add a `minimumAmount` field to their quote, right? should we add a `minimumAmount?: bigint` here? wdyt

connect/src/routes/route.ts
```diff
@@ -14,19 +16,54 @@ export type ValidationResult<OP> =
...
-// OP - Options passed to the route
-// R - Reciept of the transfer
-// Q - Quote for the transfer given the options
-export abstract class Route<N extends Network, OP, R, Q> {
+export type UnknownRoute<
+  N extends Network,
+  OP extends Options = Options,
+  R extends Receipt = Receipt,
+  Q extends Quote = Quote,
+> = Route<N, OP, R, Q>;
```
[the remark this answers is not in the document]
artursapek: or `<WO, RM, HO, LE>` lol

PR #238, opened 2024-01-20
Add check for circle message type

barnjamin:


core/definitions/src/protocols/circleBridge.ts
```diff
@@ -43,6 +43,10 @@ export namespace CircleBridge {
     attestation?: string;
   };
 
+  export const isCircleAttestation = (thing: any): thing is Attestation => {
+    return (<Attestation>thing).message !== undefined;
```
artursapek, line 47: I dont get it. how does this ensure that it's a _circle_ attestation? all it does is check that it has a `message` field. don't all of them have that?

PR #241, opened 2024-01-21
fix: holesky chainId

ConjunctiveNormalForm:
Holesky is currently defined with Optimism Sepolia's native chain id, blocking either network from being used:

`Error: Platform Evm has multiple chains with native chain id 11155420`

Grabbed chainId from https://github.com/eth-clients/holesky


barnjamin (approved):
Thanks! also checked
https://chainlist.org/chain/17000

PR #266, opened 2024-01-24
Add json serde to account for bigint/u8array types

barnjamin:


artursapek:
is this meant to be used somewhere inside the SDK?

barnjamin:
only if we have to

PR #280, opened 2024-01-26
add karura and acala configs

shunjizhan:
reference: https://evmdocs.acala.network/network/network-configuration


barnjamin (approved):
thanks!

PR #293, opened 2024-01-29
Refactor how amounts are represented

artursapek:


connect/src/protocols/tokenTransfer.ts
```diff
@@ -447,57 +455,59 @@ export class TokenTransfer<N extends Network = Network>
...
-    // but scale back to original decimals
-    const dedust = (amt: bigint) => scale(truncate(amt, srcDecimals), srcDecimals);
-    // truncate and rescale to fit destination token
-    const mintable = (amt: bigint) => scale(truncate(amt, srcDecimals), dstDecimals);
-
-    // dedust for transfer over the bridge which truncates to 8 decimals
-    let srcAmount = dedust(transfer.amount);
-    let dstAmount = mintable(transfer.amount);
+    const srcAmountTruncated = truncateAmount(srcAmount, TOKEN_BRIDGE_MAX_DECIMALS);
+    const dstAmountTruncated = scaleAmount(srcAmountTruncated, dstDecimals);
```
barnjamin, line 464: this is correct for math but nitpicking the var name, maybe call it `dstAmountMintable` or something?

PR #328, opened 2024-02-08
Governor limit check

barnjamin:


core/base/src/utils/amount.ts
```diff
@@ -151,6 +151,15 @@ export function display(amount: Amount, precision?: number): string {
   return partial.length > 0 ? `${whole}.${partial}` : whole;
 }
 
+/**
+ * Returns the decimal amount as a number (may lose precision)
+ * @param amount An Amount
+ * @returns A number, representing the decimal amount for the Amount
+ */
+export function whole(amount: Amount): number {
+  return Number(display(amount));
```
barnjamin, line 160: this seems backwards, we should create the display amt from the number

PR #342, opened 2024-02-20
Ntt payloads and better custom payload specification mechanism

nonergodic:
killed ObjectLayout item and instead moved its abilities to custom field of BytesLayout item

BytesLayoutItems can now enforce a size or a lengthSize while also natively supporting custom layouts (instead of having to wrap the layout in a CustomConversion and hence losing discrimination power, etc.)

implemented new custom payload mechanism for all payloads that have user definable payloads

also implemented NTT payloads

### DIFF
--- a/core/base/src/utils/layout/deserialize.ts
+++ b/core/base/src/utils/layout/deserialize.ts
@@ -7,17 +7,20 @@ import {
   NumSizeToPrimitive,
   NumType,
   BytesType,
-  isNumType,
-  isBytesType,
+
   numberMaxSize,
 } from "./layout";
 
-import { isLayout, checkBytesTypeEqual, checkNumEquals, isFixedBytesConversion } from "./utils";
+import {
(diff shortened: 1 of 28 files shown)


core/base/src/utils/layout/deserialize.ts
```diff
@@ -14,200 +12,196 @@ import {
...
       if (pair === undefined)
         throw new Error(`unknown id value: ${id}`);
 
       const [idOrConversionId, idLayout] = pair;
-      const [decoded, nextOffset] = internalDeserializeLayout(idLayout, encoded, newOffset);
-      return [
-        { [item.idTag ?? "id"]: hasPlainIds ? id : (idOrConversionId as any)[1],
-          ...decoded
-        },
-        nextOffset
```
barnjamin, line 209: i like that we dont have to pass these around

barnjamin (approved):

PR #347, opened 2024-02-22
Sdkv2

barnjamin:
Changes to allow a simpler install/setup process than is currently available.

Currently, every platform and protocol within that platform must be installed independently and imported/registered prior to being used.

ex:
https://github.com/wormhole-foundation/connect-sdk/blob/main/examples/src/index.ts#L1-L7

With this PR, the new `sdk` package provides all sub packages and individual platforms can be imported with the `/` path. By structuring the package this way, submodules that _arent_ imported are properly tree shaken.

In a web app ([ezui](https://github.com/barnjamin/ezui)) the bundle sizes change as we include/exclude   Platforms.

Protocols are loaded via dynamic Import which allows chunking but may be unnecessary if the protocol is unused.  We should consider alternate ways to indicate that the client actually _wants_ the protocol being loaded


sdk/package.json
```diff
@@ -0,0 +1,124 @@
...
+      "aptos": [
+        "dist/esm/aptos.d.ts"
+      ],
+      "sui": [
+        "dist/esm/sui.d.ts"
+      ]
+    }
+  },
+  "dependencies": {
+    "@wormhole-foundation/sdk-base": "0.4.4-beta.5",
```
artursapek, line 98: did you use a script to generate all of this? or are we just gonna be replace-all-ing these bad boys?

tokenRegistry/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
-  "types": "./dist/esm/index.d.ts",
+  "types": "./dist/cjs/index.d.ts",
```
artursapek, line 17: serious question... why not drop commonjs support and only use esm? Node has supported ESM for like 3 years now.

artursapek (commented):
clean af tbh, left a couple questions

sdk/package.json
barnjamin: for any _new_ packages we'd have to add them manually, for versions it will get updated with the `syncVersions` script

tokenRegistry/package.json
barnjamin: fine w/ me but not this release plzz

artursapek: fair enough lmao

artursapek (dismissed):

PR #385, opened 2024-03-07
Move payloads alongside protocols

barnjamin:


barnjamin:
opened #386 to target sdkv2 branch

PR #390, opened 2024-03-10
Create a new method to initialize the Wormhole registry

barnjamin:
Instead of initializing with the `PlatformUtil` directly in the Wormhole constructor:
```ts
const wh = new Wormhole("Testnet", [evm.Platform])
```

This pr provides an alternate method using an async function that returns the `PlatformDefinition` and awaits dynamic dependencies:
```ts
const wh = await wormhole("Testnet", [evm])
```

This method allows the required main bundle size of an anecdotal web app to shrink from 3.3mb unzipped to 2.1mb and enforces that the loading of the protocols blocks any other actions derived from the creation of the `Wormhole` object


examples/package.json
```diff
@@ -15,7 +15,7 @@
   "license": "Apache-2.0",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
-  "types": "./dist/esm/index.d.ts",
+  "types": "./dist/cjs/index.d.ts",
```
kcsongor, line 18: what's the reason for this change?

kcsongor (approved):
great patch

examples/package.json
barnjamin: matches the rest now but 

had some issue, tried setting the types to point to a diff path, not work, find and replace back, also standardize the one I missed earlier

PR #396, opened 2024-03-12
Linter

barnjamin:


connect/package.json
```diff
@@ -38,7 +38,7 @@
     "build": "npm run build:cjs && npm run build:esm",
     "rebuild": "npm run clean && npm run build",
     "clean": "rm -rf ./dist && rm -rf ./.turbo",
-    "lint": "npm run prettier && eslint --fix",
+    "lint": "npm run prettier && eslint --fix ./src --ext .ts",
```
artursapek, line 41: can you put these settings (`./src --ext .ts`) into `.eslintrc.json`?

artursapek: (and same for all the other instances of this)

artursapek: also, can we have a single top-level `.eslintrc.json` that each package uses? `../.eslintrc.json` etc

barnjamin: they all use the same eslint, will see if i can add this to the config tho

barnjamin: full 5 min of looking, didn't see how, tomorrows ben problems

artursapek: apparently its actually not possible so nvm

PR #398, opened 2024-03-13
Explicitly name all exports

barnjamin:
New script to replace any wildcard exports with exports naming everything we export


tsconfig.esm.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "module": "ES2022"
+    "module": "ES2022",
```
gabzim, line 4: doesn't this break anything? it's not valid json

barnjamin: https://json.schemastore.org/tsconfig

image

barnjamin:
closing in favor of #401

PR #410, opened 2024-03-17
Add references to platform packages so types resolve correctly

barnjamin:
Also making the platform entry points export the loader as its default so no need to import with curly braces

Without this, something like `fooAddress.toUniversalAddress().toNative("Solana")` would type to `never` since the proper types arent loaded

### DIFF
--- a/examples/src/index.ts
+++ b/examples/src/index.ts
@@ -1,5 +1,5 @@
 // EXAMPLE_IMPORTS
-import { Wormhole, wormhole } from "@wormhole-foundation/sdk";
+import { wormhole } from "@wormhole-foundation/sdk";
 import algorand from "@wormhole-foundation/sdk/algorand";
 import cosmwasm from "@wormhole-foundation/sdk/cosmwasm";
 import evm from "@wormhole-foundation/sdk/evm";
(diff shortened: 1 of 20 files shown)


examples/src/index.ts
```diff
@@ -1,14 +1,14 @@
...
-import { algorand } from "@wormhole-foundation/sdk/algorand";
-import { cosmwasm } from "@wormhole-foundation/sdk/cosmwasm";
-import { evm } from "@wormhole-foundation/sdk/evm";
-import { solana } from "@wormhole-foundation/sdk/solana";
-import { sui } from "@wormhole-foundation/sdk/sui";
+import algorand from "@wormhole-foundation/sdk/algorand";
+import cosmwasm from "@wormhole-foundation/sdk/cosmwasm";
+import evm from "@wormhole-foundation/sdk/evm";
+import solana from "@wormhole-foundation/sdk/solana";
+import sui from "@wormhole-foundation/sdk/sui";
```
barnjamin, line 7: if this is better (i think so) then the same treatment needs to be applied to the rest of the examples and the readme updater script run

PR #438, opened 2024-03-28
remove private from constructors of protocol clients

barnjamin:


connect/src/protocols/cctpTransfer.ts
```diff
@@ -64,7 +64,7 @@ export class CircleTransfer<N extends Network = Network>
 
   attestations?: AttestationReceipt<CircleTransferProtocol>[];
 
-  private constructor(
```
barnjamin, line 67: undo this

PR #441, opened 2024-03-30
Ntt learnings

barnjamin:


sdk/src/loaders.ts
```diff
@@ -0,0 +1,55 @@
...
+    return loaders[platform]!;
+  }) as PlatformLoader<Platform>[];
+
+  try {
+    // Load platforms
+    const platforms = await Promise.all(filteredLoaders.map((loader) => loader()));
+
+    // Load all protocols by default
+    await Promise.all(
+      platforms.map(
```
artursapek: you could simplify this and load all of them in parallel using `flatMap`

core/definitions/src/chain.ts
```diff
@@ -188,6 +186,28 @@ export abstract class ChainContext<
...
+  async getProtocol<PN extends ProtocolName>(
+    protocolName: ProtocolName,
+    rpc?: RpcConnection<P>,
+  ): Promise<ProtocolImplementation<P, PN>> {
+    if (!this.protocols.has(protocolName)) {
+      const ctor = this.platform.getProtocolInitializer(protocolName);
+
+      const protocol = rpc
+        ? await this.platform.getProtocol(protocolName, rpc)
+        : new ctor(this.network, this.chain, await this.getRpc(), this.config.contracts);
```
artursapek, line 204: kinda confusing. would this be valid?

```suggestion
      const protocol = await this.platform.getProtocol(protocolName, rpc ?? await this.getRpc());
```

or are these two cases not equivalent for some reason?

barnjamin: they're not equivalent, the first case will try to resolve the network/chain from the RPC, the second we pass it explicitly

PR #489, opened 2024-04-23
EVM: Don't try to connect in signer if provider already available

barnjamin:


platforms/evm/src/signer.ts
```diff
@@ -34,10 +34,18 @@ export async function getEvmSigner(
   const chain = opts?.chain ?? (await EvmPlatform.chainFromRpc(rpc))[1];
   const managedSigner = new NonceManager(signer);
 
+  if (managedSigner.provider === null) {
```
artursapek, line 37: you want to explicitly check for `null` here? it'll never be `undefined`?

barnjamin: suppose it could be but the apparent types returned are `Provider | null`

artursapek (approved):

PR #511, opened 2024-04-30
sdk-icons: init

evan-gray:
Drafted up a package to provide base64 strings of icons. Still needs to be tested and to decide if the SVGs themselves should be included as well.


core/icons/package.json
```diff
@@ -0,0 +1,72 @@
...
+      "import": {
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
+      },
+      "require": {
+        "types": "./dist/cjs/index.d.ts",
+        "default": "./dist/cjs/index.js"
+      }
+    },
+    "./chains": {
```
barnjamin, line 30: I couldn't tell what this was meant to export, maybe the stuff in `images/chains`?

core/icons/scripts/genIcons.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import svg64 from "svg64";
+import { readFileSync, writeFileSync } from "fs";
+import { chains } from "@wormhole-foundation/sdk-base";
+
+const chainToIcon: { [chain: string]: string } = {};
+
+for (const chain of chains) {
+  const svg = readFileSync(`./src/images/chains/${chain}.svg`, "utf-8");
+  const b64 = svg64(svg);
```
barnjamin, line 12: I think we can just skip this dep and use `encoding.base64.encode` on the bytes of the SVG file

I think we should also just skip the `data:image/svg+xml;base64,` prefix since it can be added by users directly or set as a const here and some `getter` to prepend it

core/icons/src/images/chains/Bsc.svg
barnjamin, line 1: is so cute

core/icons/package.json
evan-gray: I was trying to get ahead on the sparse importing (copied from base) but will just remove for now

core/icons/scripts/genIcons.ts
evan-gray: I don't think it's *quite* that simple, but I was considering just doing something similar myself to avoid the dep so I'll go ahead and do that.

PR #517, opened 2024-05-01
fix issues causing vitest to puke

artursapek:


artursapek:
might be good idea to try using `swc` to build the SDK. it's faster and I read in https://github.com/microsoft/TypeScript/issues/51783 strips out import type assertions

PR #566, opened 2024-05-24
don't throw in token bridge lookup

artursapek:
prevents a throw that happens [here](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/platforms/evm/protocols/tokenBridge/src/tokenBridge.ts#L94) from crashing routes

really this should return something like a `Result` with either `Some(token)` or `None` instead of throwing... but this will do for now :-)

### DIFF
--- a/connect/src/routes/tokenBridge/automatic.ts
+++ b/connect/src/routes/tokenBridge/automatic.ts
@@ -89,7 +89,12 @@ export class AutomaticTokenBridgeRoute<N extends Network>
     fromChain: ChainContext<N>,
     toChain: ChainContext<N>,
   ): Promise<TokenId[]> {
-    return [await TokenTransfer.lookupDestinationToken(fromChain, toChain, sourceToken)];
+    try {
+      return [await TokenTransfer.lookupDestinationToken(fromChain, toChain, sourceToken)];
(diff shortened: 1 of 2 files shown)


connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -89,7 +89,12 @@ export class AutomaticTokenBridgeRoute<N extends Network>
     fromChain: ChainContext<N>,
     toChain: ChainContext<N>,
   ): Promise<TokenId[]> {
-    return [await TokenTransfer.lookupDestinationToken(fromChain, toChain, sourceToken)];
+    try {
+      return [await TokenTransfer.lookupDestinationToken(fromChain, toChain, sourceToken)];
+    } catch (e) {
+      console.error(`Failed to get destination token: ${e}`);
```
barnjamin, line 95: I dont love the console.errors, but we need a general strat around what/when/where we console.error

artursapek: better would be to not even throw here and just return something akin to a `Result` (`Some(token) | None`)

barnjamin (approved):

PR #576, opened 2024-06-06
solana: Added client-side tx rebroadcast logic

kev1n-peters:
Implements the suggestions from here: https://docs.triton.one/chains/solana/sending-txs

### DIFF
--- a/platforms/solana/src/platform.ts
+++ b/platforms/solana/src/platform.ts
@@ -24,7 +24,9 @@ import type {
   Commitment,
   ConnectionConfig,
   ParsedAccountData,
+  RpcResponseAndContext,
   SendOptions,
+  SignatureResult,
 } from '@solana/web3.js';
 import { Connection, PublicKey } from '@solana/web3.js';
 import { SolanaAddress, SolanaZeroAddress } from './address.js';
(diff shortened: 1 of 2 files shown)


platforms/solana/src/platform.ts
```diff
@@ -176,41 +178,66 @@ export class SolanaPlatform<N extends Network>
...
+  static async sendTxWithRetry(
+    rpc: Connection,
+    tx: SignedTx,
+    retryInterval = 5000,
+  ): Promise<{
+    signature: string;
+    response: RpcResponseAndContext<SignatureResult>;
+  }> {
+    const sendOptions: SendOptions = {
+      skipPreflight: true, // TODO: need to make sure tx was simulated just before sending
```
kev1n-peters, line 206: We are skipping preflight here (important), but this is different than the previous behavior. Previously the call to `sendRawTransaction` would set `skipPreflight: false` by default. So now we either need to simulate the transaction here just before sending it or expect the caller to have done it. Callers already simulate the tx to get the compute budget, so simulating it again here would be redundant. Maybe we can add a flag to `determineComputeBudget` to throw when simulation fails. Not sure what's best here

kev1n-peters: Actually, I think we could set `skipPreflight: false` for the first call to `sendRawTransaction` so it throws if simulation fails and then set it to `true` on any subsequent calls.

platforms/solana/src/platform.ts
```diff
@@ -176,41 +178,74 @@ export class SolanaPlatform<N extends Network>
...
+    response: RpcResponseAndContext<SignatureResult>;
+  }> {
+    const commitment = sendOpts.preflightCommitment ?? rpc.commitment;
+    const signature = await rpc.sendRawTransaction(tx, {
+      ...sendOpts,
+      skipPreflight: false, // The first send should not skip preflight to catch any errors
+      maxRetries: 0,
+      preflightCommitment: commitment,
+    });
+    // TODO: Use the lastValidBlockHeight that corresponds to the blockhash used in the transaction.
```
barnjamin, line 213: should we do this TODO?

platforms/solana/src/platform.ts
```diff
@@ -176,41 +178,74 @@ export class SolanaPlatform<N extends Network>
...
+    sendOpts: SendOptions = {},
+    retryInterval = 5000,
+  ): Promise<{
+    signature: string;
+    response: RpcResponseAndContext<SignatureResult>;
+  }> {
+    const commitment = sendOpts.preflightCommitment ?? rpc.commitment;
+    const signature = await rpc.sendRawTransaction(tx, {
+      ...sendOpts,
+      skipPreflight: false, // The first send should not skip preflight to catch any errors
```
barnjamin, line 209: what is the default? assuming false, can this be removed?

kev1n-peters: Yeah it's false by default, but I would prefer to be explicit here with the accompanying comment

platforms/solana/src/platform.ts
kev1n-peters: I think we should, but there's no way to get the `lastValidBlockHeight` from a `VersionedTransaction`. I think we're going to have to change the `SignedTx` type for Solana from `Buffer` to { data: buffer, recentBlockhash, lastValidBlockHeight }`. The `recentBlockhash` and `lastValidBlockHeight` are needed in the call to `confirmTransaction`.

barnjamin (approved):

platforms/solana/src/platform.ts
kev1n-peters: Will revisit in a future PR. Not using the `recentBlockhash` and `lastValidBlockHeight` from the transaction object is how it worked prior to the changes in this PR.

PR #606, opened 2024-07-04
Add missing layouts for governance messages

M-Picco:
I noticed there were some governance messages missing from the sdk, particularly those that are not attached to common protocols (e.g. TokenBridge, WormholeRelayer, etc.). In order to add them, I had to create a few specific governance protocols under a new directory `governance`. I added tests for those I managed to find examples for.

The layouts are based on the Go sdk definitions (see [here](https://github.com/wormhole-foundation/wormhole/blob/2eb5cca8e72c5379cd444ae3f25a012c1e04ad65/sdk/vaa/payloads.go)). I also added comments with references to the contracts/source code where the messages are used, when possible.

Please let me know if there's anything I should correct, or if there's another way you'd prefer to structure the files.


core/definitions/src/protocols/governance/globalAccountant.ts
```diff
@@ -0,0 +1,39 @@
...
+export type ModificationKindEnum = 1 | 2 | 3;
+
+export const modificationKinds = [
+  [ "Add", 1 ],
+  [ "Subtract", 2 ],
+  [ "Unknown", 3 ],
+] as const satisfies MapLevel<ModificationKind, ModificationKindEnum>;
+
+export const modificationKindToEnum = constMap(modificationKinds);
+export const enumToModificationKind = constMap(modificationKinds, [1, 0]);
```
barnjamin, line 27: I think this could be sort of reversed, declaring the `modificationKinds` with types of `MapLevel<string, number>` and then deriving the types of `ModificationKind` and `ModificationKindEnum` from the constmap produced

This has the advantage of defining the values in a single place so they never get out of sync

barnjamin (approved):

PR #611, opened 2024-07-08
Missing governance layouts alt

nonergodic:
alternative branch to contrib branch - equivalent except that this one retains old VAA arg structure - only one branch should be merged


nonergodic:
This impl makes `createVAA` infeasible (because we can't look into custom conversions to determine fixed items given the current fixedDynamic impl in layout (though there might be a way to fix this with some effort)).

Hence closing since the alternative was already questionable and now is definitely not worth it.

PR #613, opened 2024-07-08
Check that the token expected from bridging is acceptable

barnjamin:


connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -90,7 +95,25 @@ export class AutomaticTokenBridgeRoute<N extends Network>
...
   ): Promise<TokenId[]> {
     try {
-      return [await TokenTransfer.lookupDestinationToken(fromChain, toChain, sourceToken)];
+      const expectedDestinationToken = await TokenTransfer.lookupDestinationToken(
+        fromChain,
+        toChain,
+        sourceToken,
+      );
+
+      const atb = await toChain.getAutomaticTokenBridge();
```
kev1n-peters, line 104: Would calling `atb.isRegisteredToken` here work?

barnjamin: good call!

kev1n-peters (approved):

PR #617, opened 2024-07-09
fix constMap get and impl subMap feature

nonergodic:
* fixed constMap.get params and return type
* implemented constMap.subMap
* fixed a few things here and there

### DIFF
--- a/core/base/src/constants/chains.ts
+++ b/core/base/src/constants/chains.ts
@@ -110,7 +110,8 @@ export const toChain = (chain: number | string | bigint): Chain => {
       if (isChainId(chain)) return chainIdToChain(chain);
       break;
     case "bigint":
-      if (isChainId(Number(chain))) return chainIdToChain(Number(chain) as ChainId);
+      if (chain < (1n << 16n) && isChainId(Number(chain)))
+        return chainIdToChain.get(Number(chain))!;
       break;
   }
   throw Error(`Cannot convert to Chain: ${chain}`);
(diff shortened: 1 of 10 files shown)


core/base/src/constants/chains.ts
```diff
@@ -110,7 +110,8 @@ export const toChain = (chain: number | string | bigint): Chain => {
       if (isChainId(chain)) return chainIdToChain(chain);
       break;
     case "bigint":
-      if (isChainId(Number(chain))) return chainIdToChain(Number(chain) as ChainId);
+      if (chain < (1n << 16n) && isChainId(Number(chain)))
+        return chainIdToChain.get(Number(chain))!;
```
barnjamin, line 114: yt we should add a `encoding.bignum.downcast` or something that checks for safety first?

barnjamin (dismissed):

core/base/src/utils/array.ts
```diff
@@ -1,15 +1,35 @@
...
+export const concat = (...bytes: RoArray<Uint8Array>): Uint8Array => {
+  const len = bytes.reduce((acc, b) => acc + b.length, 0);
+  const result = new Uint8Array(len);
+  let offset = 0;
+  for (const b of bytes) {
+    result.set(b, offset);
+    offset += b.length;
+  }
+  return result;
+}
```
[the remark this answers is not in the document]
nonergodic: Awesome. Putting it in encoding is a much better place for it too. I should have thought of it myself. I put it in array.ts essentially entirely for lexicographical reasons and I shall delete this silly duplicate posthaste!

core/base/src/constants/chains.ts
nonergodic: Yeah, I think that's a great idea too!

barnjamin (approved):

PR #618, opened 2024-07-09
hash aptos addresses when searching for foreign asset

artursapek:


platforms/aptos/src/address.ts
```diff
@@ -1,5 +1,5 @@
 import type { Address } from "@wormhole-foundation/sdk-connect";
-import { UniversalAddress, encoding, registerNative } from "@wormhole-foundation/sdk-connect";
+import { UniversalAddress, registerNative, encoding } from "@wormhole-foundation/sdk-connect";
```
barnjamin, line 2: Can these be

barnjamin (approved):

PR #625, opened 2024-07-12
connect: Added DestinationQueued transfer state

kev1n-peters:
This state may occur after the transaction has been submitted on the destination chain (DestinationInitiated) and it has been queued for some reason (e.g. NTT inbound rate limited).


connect/src/types.ts
```diff
@@ -68,6 +69,18 @@ export interface RedeemedTransferReceipt<AT, SC extends Chain = Chain, DC extend
...
+export interface DestinationQueuedTransferReceipt<
+  AT,
+  SC extends Chain = Chain,
+  DC extends Chain = Chain,
+> extends BaseTransferReceipt<SC, DC> {
+  state: TransferState.DestinationQueued;
+  originTxs: TransactionId<SC>[];
+  attestation: Required<AT>;
+  destinationTxs?: TransactionId<DC>[];
+  queueExpirationTime: Date;
```
artursapek, line 81: "expiration" might be a confusing term here... it imply the transfer is no longer valid or eligible. what do you think about `queueReleaseTime`? it's the time the transfer will be released from the queue.

kev1n-peters: Changed

artursapek (approved):

PR #629, opened 2024-07-15
Manual resume

artursapek:
This adds a required `resume` method to `ManualRoute`, which takes a `TransactionId` (combination of chain + hash) and returns a `Receipt`.

### DIFF
--- a/connect/src/protocols/cctp/cctpTransfer.ts
+++ b/connect/src/protocols/cctp/cctpTransfer.ts
@@ -209,7 +209,7 @@ export class CircleTransfer<N extends Network = Network>
   }
 
   // init from source tx hash
-  static async fromTransaction<N extends Network>(
+  private static async fromTransaction<N extends Network>(
     wh: Wormhole<N>,
     from: TransactionId,
     timeout: number,
--- a/connect/src/protocols/cctp/cctpTransfer.ts
+++ b/connect/src/protocols/cctp/cctpTransfer.ts
@@ -209,7 +209,7 @@ export class CircleTransfer<N extends Network = Network>
   }
 
   // init from source tx hash
-  private static async fromTransaction<N extends Network>(
+  static async fromTransaction<N extends Network>(
     wh: Wormhole<N>,
     from: TransactionId,
     timeout: number,
--- a/connect/src/protocols/tokenBridge/tokenTransfer.ts
+++ b/connect/src/protocols/tokenBridge/tokenTransfer.ts
@@ -192,7 +192,7 @@ export class TokenTransfer<N extends Network = Network>
     return tt;
   }
 
-  private static async fromTransaction<N extends Network>(
+  static async fromTransaction<N extends Network>(
     wh: Wormhole<N>,
     from: TransactionId,
     timeout: number,
--- a/connect/src/protocols/tokenBridge/tokenTransfer.ts
+++ b/connect/src/protocols/tokenBridge/tokenTransfer.ts
@@ -405,10 +405,8 @@ export namespace TokenTransfer {
         throw new Error('no');
       }
 
-      const ben = wh.getChain(receipt.attestation.attestation.payload.to.chain);
-
       let isComplete = await TokenTransfer.isTransferComplete(
-        ben,
+        wh.getChain(receipt.attestation.attestation.payload.to.chain),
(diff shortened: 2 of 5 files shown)


connect/src/protocols/tokenBridge/tokenTransfer.ts
```diff
@@ -343,13 +343,11 @@ export namespace TokenTransfer {
     receipt: TokenTransfer.TransferReceipt<SC, DC>,
     timeout: number = DEFAULT_TASK_TIMEOUT,
     fromChain?: ChainContext<N, SC>,
-    toChain?: ChainContext<N, DC>,
```
barnjamin, line 346: why take this out? if anyone is providing it, they'll be sad

connect/src/routes/tokenBridge/manual.ts
```diff
@@ -168,6 +170,21 @@ export class TokenBridgeRoute<N extends Network>
       destinationTxs: dstTxIds,
     };
   }
+  
+  async resume(tx: TransactionId): Promise<R> {
+    const attestation = await TokenTransfer.getTransferVaa(this.wh, tx.txid);
+    const id = await TokenTransfer.getTransferMessage(this.wh.getChain(tx.chain), tx.txid);
```
barnjamin, line 176: if this gets `resume`'d before the vaa is signed, does this fail? maybe use the same method as the cctp with `TokenTransfer.from` + a method to produce a receipt

connect/src/routes/route.ts
```diff
@@ -130,6 +130,7 @@ export abstract class ManualRoute<
   NATIVE_GAS_DROPOFF_SUPPORTED = false;
   IS_AUTOMATIC = false;
   public abstract complete(sender: Signer, receipt: R): Promise<R>;
+  public abstract resume(tx: TransactionId): Promise<R>;
```
kev1n-peters, line 133: Should this go on the base class `Route` since we may want to `resume` a transaction for any route? For example, an auto NTT transfer may need to be manually resumed to complete the transfer if it was queued on the destination chain.

connect/src/routes/tokenBridge/manual.ts
artursapek: it fails and its ok that it does for now

connect/src/routes/route.ts
artursapek: will make resume-able a new abstract class

connect/src/protocols/tokenBridge/tokenTransfer.ts
artursapek: i took it out because we don't require it. we should always be able to get it from the attestation. so why let people pass it in, what does that do for them? there's a chance they pass in the wrong chain and it gets fucky

connect/src/routes/route.ts
artursapek: (decided to keep this as-is)

connect/src/protocols/tokenBridge/tokenTransfer.ts
artursapek: just looked at this again and made that change actually :+1:
thanks

connect/src/routes/tokenBridge/manual.ts
artursapek: just looked at this again and made that change actually 👍
thanks

connect/src/protocols/tokenBridge/tokenTransfer.ts
barnjamin: It provides the ability to re-use the same chain in the case that its already created with a specific RPC

connect/src/protocols/cctp/cctpTransfer.ts
```diff
@@ -209,7 +209,7 @@ export class CircleTransfer<N extends Network = Network>
   }
 
   // init from source tx hash
-  private static async fromTransaction<N extends Network>(
+  static async fromTransaction<N extends Network>(
```
barnjamin, line 212: why make these public and not just call the `from` method which should invoke this one?

artursapek: just made this change :+1:

barnjamin (approved):

PR #630, opened 2024-07-17
ran turbo linter

evgeniko:
ran "npm run lint", without any other chances


platforms/solana/protocols/core/src/postMessageLayout.ts
```diff
@@ -1,6 +1,6 @@
-import type { Layout } from '@wormhole-foundation/sdk-connect';
+import type { Layout ,
+  LayoutToType} from '@wormhole-foundation/sdk-connect';
```
barnjamin, line 2: same

barnjamin (approved):

PR #640, opened 2024-07-25
connect: auto tb gas dropoff max amount fix

kev1n-peters:
The max swap amount should be the input amount if it's lower than the max swap amount returned by the destination relayer contract.

### DIFF
--- a/connect/src/routes/tokenBridge/automatic.ts
+++ b/connect/src/routes/tokenBridge/automatic.ts
@@ -181,10 +181,23 @@ export class AutomaticTokenBridgeRoute<N extends Network>
     if (params.options && params.options.nativeGas > 0) {
       const dtb = await request.toChain.getAutomaticTokenBridge();
       // the maxSwapAmount is in destination chain decimals
-      const maxSwapAmount = await dtb.maxSwapAmount(request.destination.id.address);
+      let maxSwapAmount = await dtb.maxSwapAmount(request.destination.id.address);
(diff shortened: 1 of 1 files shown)


connect/src/routes/tokenBridge/automatic.ts
```diff
@@ -181,10 +181,23 @@ export class AutomaticTokenBridgeRoute<N extends Network>
...
+      );
+      const dstDecimals = await request.toChain.getDecimals(request.destination.id.address);
+      const dstAmountReceivable = amount.units(
+        amount.scale(redeemableAmountTruncated, dstDecimals),
+      );
+      if (dstAmountReceivable < maxSwapAmount) {
+        // can't swap more than the receivable amount
+        maxSwapAmount = dstAmountReceivable;
+      }
       const scale = 10000;
```
evgeniko, line 197:
```suggestion
      const GAS_SCALE = 10000;
```

evgeniko (approved):

connect/src/routes/tokenBridge/automatic.ts
kev1n-peters: I agree that this name is more descriptive and we generally use uppercase for constants (so this suggestion is definitely an improvement), but I'm going to leave it as is for now so not add more changes to the PR that aren't absolutely necessary. Thank you for the feedback though, maybe we can revisit this

PR #645, opened 2024-07-29
connect: Added InReview transfer state, governor warnings and fixes

kev1n-peters:
The InReview transfer state is currently set when a token bridge transfer is enqueued in the Governor.

Added Governor warnings to token bridge quotes.

Fixed an issue where the token price wasn't being looked up by its origin chain.


emreboga (approved):
LGTM

PR #646, opened 2024-07-30
connect: Added eta to quote interface

kev1n-peters:
The eta tells us the estimated transfer time in milliseconds (essentially finality time).

### DIFF
--- a/connect/src/protocols/cctp/cctpTransfer.ts
+++ b/connect/src/protocols/cctp/cctpTransfer.ts
@@ -37,6 +37,7 @@ import type {
 import { TransferState, isAttested, isSourceFinalized, isSourceInitiated } from "../../types.js";
 import { Wormhole } from "../../wormhole.js";
 import type { WormholeTransfer } from "../wormholeTransfer.js";
+import { finality } from "@wormhole-foundation/sdk-base";
 
 export class CircleTransfer<N extends Network = Network>
   implements WormholeTransfer<CircleTransfer.Protocol>
@@ -225,10 +226,10 @@ export class CircleTransfer<N extends Network = Network>
     try {
       msgIds = await fromChain.parseTransaction(txid);
     } catch (e: any) {
-      if (e.message.includes('no bridge messages found')) {
+      if (e.message.includes("no bridge messages found")) {
         // This means it's a Circle attestation; swallow
       } else {
-        throw e
+        throw e;
       }
     }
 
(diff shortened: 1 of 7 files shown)


connect/src/protocols/cctp/cctpTransfer.ts
```diff
@@ -584,10 +585,14 @@ export namespace CircleTransfer {
     const dstToken = Wormhole.chainAddress(dstChain.chain, dstUsdcAddress);
     const srcToken = Wormhole.chainAddress(srcChain.chain, srcUsdcAddress);
 
+    // https://developers.circle.com/stablecoins/docs/required-block-confirmations
+    const eta =
+      srcChain.chain === "Polygon" ? 2_000 * 200 : finality.estimateFinalityTime(srcChain.chain);
```
emreboga, line 590: @kev1n-peters Is this going to be a fixed ETA when it's Polygon?

emreboga: I mean it says 200 blocks in the linked document, but is it 2 secs per block as fixed time in Polygon?

kev1n-peters: Yeah, it's always going to be a fixed time (`estimateFinalityTime` also returns a "fixed time", but we are overriding the finality time for Polygon here since it's different for CCTP transfers than for token bridge transfers

connect/src/routes/request.ts
```diff
@@ -68,6 +68,8 @@ export class RouteTransferRequest<N extends Network> {
       dq.warnings = [...quote.warnings];
     }
 
+    dq.eta = quote.eta;
```
emreboga, line 71: All the above props are set after empty checks. Do you think we need the same check for `eta`, or not necessary?

core/base/src/constants/finality.ts
```diff
@@ -176,3 +176,18 @@ export function consistencyLevelToBlock(
...
+ * @returns The estimated time in milliseconds
+ */
+export function estimateFinalityTime(chain: Chain): number {
+  const finality = finalityThreshold.get(chain);
+  if (finality === undefined) throw new Error("Cannot find finality for " + chain);
+
+  const time = blockTime.get(chain);
+  if (time === undefined) throw new Error("Cannot find block time for " + chain);
+
+  return finality * time;
```
emreboga, line 192: Do we expect both of these being number types? Are the types from `finalityThreshold` and `blockTime` being `undefined` or a `number`?

connect/src/routes/request.ts
kev1n-peters: I don't think so since quote.eta will either be a number or undefined, and if it's undefined then dq.eta will also be undefined. I don't think it's necessary unless you think we should change it

core/base/src/constants/finality.ts
kev1n-peters: They will both be number types. If they are not found for the given chain (undefined) then we throw an error

connect/src/routes/request.ts
emreboga: I don't have a strong opinion about it, more for consistency as other props are set after a check. It's your call 👍

PR #649, opened 2024-07-31
tokenBridge protocol: add token address conversion methods

kev1n-peters:
- Added getTokenUniversalAddress and getTokenNativeAddress methods to each platform's token bridge protocol implementation.
- Some chains like Aptos and Sui require fetching on-chain data for token address conversions (universal to native and vice versa).
- Fixed issue where lookupDestinationToken would return a universal address, causing issues in functions expecting a native address.
- lookupDestinationToken now consistently returns a native address.
- Resolved issues with transferring native Sui and Aptos tokens back to their origin chains.

### DIFF
--- a/connect/src/protocols/tokenBridge/tokenTransfer.ts
+++ b/connect/src/protocols/tokenBridge/tokenTransfer.ts
@@ -522,7 +522,8 @@ export namespace TokenTransfer {
           address = token.address;
         }
         lookup = await tb.getOriginalAsset(address);
-      } catch (e) {
+      } catch (e: any) {
+        if (!e.message.includes("not a wrapped asset")) throw e;
         // not a from-chain native wormhole-wrapped one
         let address: NativeAddress<SC>;
         if (UniversalAddress.instanceof(token.address)) {
@@ -651,7 +652,8 @@ export namespace TokenTransfer {
     } else if (UniversalAddress.instanceof(transfer.token.address)) {
       try {
         srcToken = (await srcTb.getWrappedAsset(transfer.token)) as NativeAddress<Chain>;
-      } catch (e) {
+      } catch (e: any) {
+        if (!e.message.includes("not a wrapped asset")) throw e;
         srcToken = await srcTb.getTokenNativeAddress(srcChain.chain, transfer.token.address);
       }
     } else {
--- a/connect/src/protocols/tokenBridge/tokenTransfer.ts
+++ b/connect/src/protocols/tokenBridge/tokenTransfer.ts
@@ -5,6 +5,7 @@ import type {
   AutomaticTokenBridge,
   ChainContext,
   Signer,
+  NativeAddress,
   TokenId,
   TokenTransferDetails,
   TransactionId,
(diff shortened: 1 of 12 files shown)


connect/src/wormhole.ts
```diff
@@ -222,6 +223,38 @@ export class Wormhole<N extends Network> {
     return await tb.getOriginalAsset(token.address);
   }
 
+  /**
+   * Returns the UniversalAddress of the token. This may require fetching on-chain data.
+   * @param chain The chain to get the UniversalAddress for
+   * @param token The address to get the UniversalAddress for
+   * @returns The UniversalAddress of the token
+   */
+  async getTokenUniversalAddress<C extends Chain>(
```
emreboga, line 232: Is this function exported or maybe in the future? If so, we might want to check the params.

connect/src/wormhole.ts
```diff
@@ -222,6 +223,38 @@ export class Wormhole<N extends Network> {
...
+  }
+
+  /**
+   * Returns the native address of the token. This may require fetching on-chain data.
+   * @param chain The chain to get the native address for
+   * @param originChain The chain the token is from / native to
+   * @param token The address to get the native address for
+   * @returns The native address of the token
+   */
+  async getTokenNativeAddress<C extends Chain>(
```
emreboga, line 248: Same as above.

connect/src/protocols/tokenBridge/tokenTransfer.ts
```diff
@@ -627,14 +644,26 @@ export namespace TokenTransfer {
...
-    const srcDecimals = await srcChain.getDecimals(transfer.token.address);
+    const srcTb = await srcChain.getTokenBridge();
+    let srcToken: NativeAddress<Chain>;
+    if (isNative(transfer.token.address)) {
+      srcToken = await srcTb.getWrappedNative();
+    } else if (UniversalAddress.instanceof(transfer.token.address)) {
+      try {
+        srcToken = (await srcTb.getWrappedAsset(transfer.token)) as NativeAddress<Chain>;
+      } catch (e) {
+        srcToken = await srcTb.getTokenNativeAddress(srcChain.chain, transfer.token.address);
```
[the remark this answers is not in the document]
kev1n-peters: I noticed in some other places that we don't check the exception, but I'm going to fix that. We should check that the exception being thrown is what we expect otherwise re-raise it.

connect/src/wormhole.ts
kev1n-peters: I'm not sure what you mean by checking the params, do you have an example?

emreboga: I meant null-checking the params inside the function, but we may also let it throw so the caller can handle it.

kev1n-peters: I think for now the simplest thing is to let the caller catch the exception or let it bubble up in case of an error. If there's a network error, then there's nothing we can really do

emreboga (approved):
LGTM!

PR #650, opened 2024-08-01
examples: fix type imports

kev1n-peters:


kev1n-peters:
This PR just refactors imports in the examples so that types are imported via `import type`

nonergodic (approved):
lgtm

PR #652, opened 2024-08-02
cctp: gas drop-off fix

kev1n-peters:
- use wormhole circle relayer instead of token bridge relayer to compute max swap and gas drop-off amounts
- the connect interface specifies a percentage that is applied to the the max swap amount to get the gas drop-off amount

### DIFF
--- a/connect/src/routes/cctp/automatic.ts
+++ b/connect/src/routes/cctp/automatic.ts
@@ -168,7 +168,7 @@ export class AutomaticCCTPRoute<N extends Network>
       );
     }
 
-    const transferableAmount = amount.units(amt) - fee;
+    const redeemableAmount = amount.units(amt) - fee;
 
     const options = params.options ?? this.getDefaultOptions();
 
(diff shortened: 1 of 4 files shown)


connect/src/routes/cctp/automatic.ts
```diff
@@ -179,9 +179,21 @@ export class AutomaticCCTPRoute<N extends Network>
     let nativeGasAmount = 0n;
 
     if (nativeGasPerc > 0.0) {
+      const dcb = await request.toChain.getAutomaticCircleBridge();
+      let maxSwapAmount = await dcb.maxSwapAmount();
```
emreboga, line 183: Can `getAutomaticCircleBridge()` return undefined? Even if there is a slight possibility, it'd be good to check 👍

connect/src/routes/cctp/automatic.ts
```diff
@@ -179,9 +179,21 @@ export class AutomaticCCTPRoute<N extends Network>
...
+
       const scale = 10000;
-      const scaledGas = BigInt(nativeGasPerc * scale);
-      nativeGasAmount = (transferableAmount * scaledGas) / BigInt(scale);
+      const scaledGas = BigInt(Math.floor(nativeGasPerc * scale));
+      // the native gas percentage is applied to the max swap amount
+      nativeGasAmount = (maxSwapAmount * scaledGas) / BigInt(scale);
+      if (nativeGasAmount === redeemableAmount && nativeGasAmount > 0n) {
+        // edge case: transfer will revert if the native gas amount is equal to the redeemable amount
+        nativeGasAmount -= 1n;
```
emreboga, line 195: What is the significance of subtracting 1 bigint? Just to make it below zero?

kev1n-peters: If the USDC amount we're swapping for native gas is equal to the amount we're transferring minus the relayer fee, then the contract will revert, so we need to make sure it's less than that in the rare case that it's equal

connect/src/routes/cctp/automatic.ts
kev1n-peters: The function to get it will throw if it's undefined and I'm following the pattern used across the SDK here for getting the chain's protocol implementation

emreboga (approved):
LGTM!

PR #665, opened 2024-08-08
solana: fix determinePriorityFee method for VersionedTransactions

kev1n-peters:
We were not resolving the address lookup table accounts


artursapek (approved):
<3

PR #666, opened 2024-08-08
0.10.0 version bump

kev1n-peters:


artursapek:
so i broke interfaces with my static properties change. i think maybe 0.10.0 appropriate here?

artursapek:
also
image

artursapek (approved):

PR #667, opened 2024-08-09
solana: added rpc-websockets dependency

kev1n-peters:
Fixes this issue: https://stackoverflow.com/questions/78566652/solana-web3-js-cannot-find-module-rpc-websockets-dist-lib-client


evgeniko (approved):
ty!

PR #682, opened 2024-08-27
tests: fix finality tests, run CI tests on pull requests

kev1n-peters:
running into this issue: https://github.com/jestjs/jest/issues/11617


artursapek (approved):
yaml is so weird

PR #687, opened 2024-09-04
Remove `isAvailable` method from Route interface

artursapek:


kev1n-peters (approved):

kev1n-peters:
Looking at the NTT SDK:
https://github.com/wormhole-foundation/example-native-token-transfers/blob/3c626bd7bd8b430d62b4273b75c5729bdc85ddfe/sdk/route/src/automatic.ts#L100-L111

It seems like this logic should be moved into the `isProtocolSupported` method? The only issue is that we'd have to make that method `async`.

PR #709, opened 2024-10-01
Exports token constants in `base` package.

scnale:
And also converts all comments on functions and fields to docstrings.


scnale:
It's not clear to me that this really helps so I'm closing it for now.

PR #710, opened 2024-10-05
fix, cleanup, and tighten type programming utils

nonergodic:
The branch name is outdated. Initially intended to only add the one function that I needed, but then realized various issues in the code (primarily the usage of RoArray, when Tuple types were expected and "generic" RoArray that were not tuples under the hood actually produced incorrect results).

Types are now handled more appropriately and robustly.
e.g.
```
const arr = [[1,2], [3,4]];
const zipped = zip(arr); //previously had type never[], now has type RoArray2D<number>
```

The main downside is that compile time has gone up significantly (by 15 seconds, from 55s to 70s on my machine).

### DIFF
--- a/core/base/src/constants/guardians.ts
+++ b/core/base/src/constants/guardians.ts
@@ -1,5 +1,5 @@
 import type { MapLevels} from './../utils/index.js';
-import { constMap, column, cartesianRightRecursive } from './../utils/index.js';
+import { constMap, filterIndexes, zip, cartesianRightRecursive } from './../utils/index.js';
 import type { Network } from './networks.js';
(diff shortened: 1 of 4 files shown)


artursapek:
this dude's frontal lobe bigger than my entire brain

nonergodic:
> this dude's frontal lobe bigger than my entire brain

[Yes, it's all very impressive.](https://briannadon.com/wp-content/uploads/2023/03/a-common-sense-approach-to-performance-appraisals-2014-02-25-1229971990-e1677886739392.png?w=600&h=190&crop=1)

PR #720, opened 2024-10-17
removed automatic route isAvailable method

kev1n-peters:


connect/src/routes/resolver.ts
```diff
@@ -102,25 +100,6 @@ export class RouteResolver<N extends Network> {
...
-    // Next, we make sure all supported routes are available. For relayed routes, this will ping
-    // the relayer to make sure it's online.
-    return await Promise.all(
-      supportedRoutes.map(
-        async (
-          rc,
-        ): Promise<[Route<N, Options, ValidatedTransferParams<Options>, Receipt>, boolean]> => {
-          const route = new rc(this.wh);
-          try {
-            const available = isAutomatic(route) ? await route.isAvailable(request) : true;
```
kev1n-peters, line 114: Calling `route.isAvailable` here is not necessary since the only route that actually returns something other than `true` is the AutomaticTokenBridge route and it performs this same check when calling `supportedSourceTokens`.

artursapek (approved):
excellent

PR #724, opened 2024-10-23
portico: fetch swap event

kev1n-peters:
In addition to checking if the transfer was completed, we now check if the swap actually succeeded or failed. When the swap fails, the Wormhole-wrapped / highway token is received instead. If we can't find the swap event, then we assume it succeeded.

### DIFF
--- a/platforms/evm/protocols/portico/src/bridge.ts
+++ b/platforms/evm/protocols/portico/src/bridge.ts
@@ -10,7 +10,9 @@ import type {
 } from '@wormhole-foundation/sdk-connect';
 import {
   PorticoBridge,
+  TokenTransfer,
   Wormhole,
+  amount,
   canonicalAddress,
   isEqualCaseInsensitive,
   nativeChainIds,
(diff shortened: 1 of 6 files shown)


platforms/evm/protocols/portico/src/bridge.ts
```diff
@@ -349,4 +350,87 @@ export class EvmPorticoBridge<
...
+    // NOTE: If we can't find the event, assume the swap succeeded
+    const tokenBridge = this.tokenBridge.tokenBridge;
+    const filter = tokenBridge.filters.TransferRedeemed(
+      toChainId(vaa.emitterChain),
+      vaa.emitterAddress.toString(),
+      vaa.sequence,
+    );
+
+    // Search for the event in the last 15 minutes
+    const latestBlock = await EvmPlatform.getLatestBlock(this.provider);
```
kev1n-peters, line 396: We have to constrain the range we're searching for otherwise the RPC may return an error

PR #726, opened 2024-10-24
Add Monad Devnet support

bruce-riley:
This PR adds support for the Monad Devnet environment. This is being added separately from regular Monad testnet (TBD) because it is expected that this environment will be upgraded and reset more often than a standard testnet environment.

This PR only adds support for the core contract, which is available [here](https://brightstar-884.devnet1.monad.xyz/address/0x376428e7f26D5867e69201b275553C45B09EE090).

This PR does not specify an RPC endpoint because there is not currently a public one available.


core/base/src/constants/finality.ts
```diff
@@ -76,6 +76,7 @@ const finalityThresholds = [
   ["BaseSepolia", 512],
   ["OptimismSepolia", 512],
   ["PolygonSepolia", 32],
+  ["MonadDevnet", 0],
```
evan-gray, line 79: should this be a 1 like Fantom?

core/base/src/constants/rpc.ts
```diff
@@ -91,6 +91,7 @@ const rpcConfig = [[
     ["Snaxchain",       "https://testnet.snaxchain.io"],
     ["Unichain",        "https://sepolia.unichain.org"],
     ["Worldchain",      "https://worldchain-sepolia.g.alchemy.com/public"],
+    ["MonadDevnet",     "TBD"],
```
evan-gray, line 94: oof. what would @kev1n-peters recommend here?

kev1n-peters: Maybe we could set it to the empty string and add a comment saying, "TODO: no public rpc is currently available, override with a custom rpc for now"

core/base/src/constants/finality.ts
bruce-riley: I changed MonadDevent and Berachain to 1 and SnaxChain, Unichain and Worldchain to 512 per offline discussion with @evan-gray.

evan-gray: concur, this is just to match the other OP Stack chains, not because it's factually accurate 👀

evan-gray (approved):

kev1n-peters (approved):

PR #728, opened 2024-10-28
Include Spl Token 2022 in getBalances

yuli-ferna:
In LibertAI token NTT integration we saw this bug in Solana token selector:

Captura de pantalla 2024-10-28 a la(s) 3 24 13 p m

It shows as disabled when I have a balance in my wallet. It happens because the LibertAI token uses the program [spl token 2022](https://github.com/solana-labs/solana-program-library/blob/21f0430147686d2a948e090b358b1fbfe2d4bb03/token/js/src/constants.ts#L7), and the SDK only queries the spl token program https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/platforms/solana/src/platform.ts#L153-L158.

The solution we found is to include request with the spl token 2022 program id https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/constants.ts#L7

### DIFF
--- a/platforms/solana/src/platform.ts
+++ b/platforms/solana/src/platform.ts
@@ -19,7 +19,7 @@ import {
 } from '@wormhole-foundation/sdk-connect';
 import { SolanaChain } from './chain.js';
 
-import { TOKEN_PROGRAM_ID } from '@solana/spl-token';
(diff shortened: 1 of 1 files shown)


solanoepalacio (approved):

artursapek (approved):

solanoepalacio (approved):

solanoepalacio:
@artursapek @nonergodic could you merge this one when possible? thanks

PR #738, opened 2024-11-13
extract layouting into standalone package and adjust for changes

nonergodic:
This is the mostly backwards compatible version that reverts the naming to the current TS SDK names.

It's mostly but not fully backwards compatible because the package exports fewer types, so this is somewhat of a breaking change (though I doubt most integrators will be affected).

The standalone package:
https://www.npmjs.com/package/binary-layout


kev1n-peters:
Smoke tested these changes in Connect with https://github.com/wormhole-foundation/wormhole-sdk-ts/releases/tag/1.1.2-alpha.0 insatlled and didn't encounter any issues.

kev1n-peters (approved):

PR #742, opened 2024-11-19
solana-core: export secp256k1, add optional guardianSetData param

kev1n-peters:


core/base/src/constants/guardians.ts
```diff
@@ -1,40 +1,78 @@
...
-    ["0x178e21ad2E77AE06711549CFBB1f9c7a9d8096e8", "Syncnode"         ],
-    ["0x5E1487F35515d02A92753504a8D75471b9f49EdB", "Triton"           ],
-    ["0x6FbEBc898F403E4773E95feB15E80C9A99c8348d", "StakingFacilities"],
-  ]], [
-  "Testnet", [
-    ["0x13947Bd48b18E53fdAeEe77F3473391aC727C638", "Testnet guardian"]
-  ]]
-] as const satisfies MapLevels<[Network, string, string]>;
-
-export const [guardianKeys, guardianNames] =
```
kev1n-peters, line 33: I'm not sure if we should keep these and create them using just the latest guardian set

nonergodic:
I think the real answer here is adding a `typedMap` implementation that is similar to `constMap`, but instead of using literal types, it should use the widened types. It could have exactly the same functionality as `constMap` only that its `()` operator throws when a lookup fails, while `.get` returns the widened type or `undefined`.

But that's ofc a bit larger of an undertaking.

PR #746, opened 2024-11-27
Fix infinite loop in address parsing

artursapek:
A Backpack dev ran into a bug where he was parsing a Sui address with `Ethereum` as a chain by mistake, and it caused this function to call `toNative` infinitely.

This change maintains the fallback behavior where it tries to treat unparsed `UniversalAddress` values as such, but without the potential to loop infinitely.

image

image


kev1n-peters:
Nice fix, ran into this before

kev1n-peters (approved):

PR #754, opened 2024-12-11
SR Addresses: Mantle, XLayer, Worldchain, Monad-Devnet

hernandp:


core/base/src/constants/contracts/relayer.ts
```diff
@@ -35,7 +38,8 @@ export const relayerContracts = [[
     ["Sepolia",         "0x7B1bD7a6b4E61c2a123AC6BC2cbfC614437D0470"],
     ["ArbitrumSepolia", "0x7B1bD7a6b4E61c2a123AC6BC2cbfC614437D0470"],
     ["BaseSepolia",     "0x93BAD53DDfB6132b0aC8E37f6029163E63372cEE"],
-    ["OptimismSepolia", "0x93BAD53DDfB6132b0aC8E37f6029163E63372cEE"]
+    ["OptimismSepolia", "0x93BAD53DDfB6132b0aC8E37f6029163E63372cEE"],
+    ["Monad",           "0x362fca37E45fe1096b42021b543f462D49a5C8df"]
```
tonyjin: Don't think we need to include this - this is a temporary devnet connection that we expect to update to their testnet deployment early in Q1

kev1n-peters (approved):

PR #781, opened 2025-01-17
Fix WSOL redemption

artursapek:
Addresses https://github.com/wormhole-foundation/wormhole-connect/issues/3206

### DIFF
--- a/connect/src/protocols/tokenBridge/tokenTransfer.ts
+++ b/connect/src/protocols/tokenBridge/tokenTransfer.ts
@@ -25,6 +25,7 @@ import {
   canonicalAddress,
   deserialize,
   isNative,
+  isSameToken,
   isTokenId,
   isTokenTransferDetails,
   isTransactionIdentifier,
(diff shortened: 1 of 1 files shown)


connect/src/protocols/tokenBridge/tokenTransfer.ts
```diff
@@ -860,9 +860,9 @@ export namespace TokenTransfer {
         dstChain,
         _transfer.token,
       );
-      // TODO: If the token is native, no need to overwrite the destination address check for native
-      //if (!destinationToken.address.equals((await dstChain.getNativeWrappedTokenId()).address))
-      _transfer.to = await dstChain.getTokenAccount(_transfer.to.address, destinationToken.address);
+      if (!isNative(destinationToken.address)) {
+        _transfer.to = await dstChain.getTokenAccount(_transfer.to.address, destinationToken.address);
```
kev1n-peters, line 864: If `destinationToken.address` is the `native` string literal, then we won't  override the `to` address (i.e. the associated token account). I think prior to this change https://github.com/wormhole-foundation/wormhole-sdk-ts/commit/e824c8fd32591914fb5125b9194bee8f1c46f08e the `native` string literal would never be passed, but now it is. I think what we want to do here is this:
```
if (isNative(destinationToken.address)) {
  const nativeWrappedTokenId = await dstChain.getNativeWrappedTokenId();
  _transfer.to = await dstChain.getTokenAccount(_transfer.to.address, nativeWrappedTokenId.address);
} else {
  _transfer.to = await dstChain.getTokenAccount(_transfer.to.address, destinationToken.address)
}
```

kev1n-peters (approved):

kev1n-peters (approved):

PR #785, opened 2025-01-19
Added Triton One improved priority fee API support

kev1n-peters:
Added determinePriorityFeeTritonOne function

### DIFF
--- a/core/base/src/utils/array.ts
+++ b/core/base/src/utils/array.ts
@@ -174,3 +174,31 @@ export type Cartesian<L, R> =
   : R extends RoArray
   ? [...{ [K in keyof R]: K extends `${number}` ? [L, R[K]] : never }]
   : [L, R];
+
+export const median = <T extends number | bigint>(arr: T[], isSorted: boolean = false): T => {
+  if (arr.length === 0) throw new Error("Can't calculate median of empty array");
+
+  const sorted = isSorted
+    ? arr
+    : [...arr].sort((a, b) => {
+        // handle bigint and number
+        return a > b ? 1 : a < b ? -1 : 0;
+      });
+
+  const mid = Math.floor(sorted.length / 2);
+
+  if (sorted.length % 2 === 1) {
+    return sorted[mid]!;
+  }
+
+  const left = sorted[mid - 1]!;
+  const right = sorted[mid]!;
+
+  if (typeof left === "bigint" && typeof right === "bigint") {
+    return ((left + right) / 2n) as T;
+  } else if (typeof left === "number" && typeof right === "number") {
+    return ((left + right) / 2) as T;
+  } else {
+    throw new Error("Can't calculate median of array with mixed number and bigint");
+  }
+};
--- a/core/base/src/utils/array.ts
+++ b/core/base/src/utils/array.ts
@@ -175,9 +175,9 @@ export type Cartesian<L, R> =
   ? [...{ [K in keyof R]: K extends `${number}` ? [L, R[K]] : never }]
   : [L, R];
 
-export function median(arr: number[], isSorted?: boolean): number;
-export function median(arr: bigint[], isSorted?: boolean): bigint;
-export function median(arr: (number | bigint)[], isSorted: boolean = false): number | bigint {
+export function median(arr: RoArray<number>, isSorted?: boolean): number;
+export function median(arr: RoArray<bigint>, isSorted?: boolean): bigint;
+export function median(arr: RoArray<number> | RoArray<bigint>, isSorted: boolean = false): number | bigint {
   if (arr.length === 0) throw new Error("Can't calculate median of empty array");
 
   const sorted = isSorted ? arr : [...arr].sort((a, b) => (a > b ? 1 : a < b ? -1 : 0)); // handle bigint and number
--- a/core/base/src/utils/array.ts
+++ b/core/base/src/utils/array.ts
@@ -175,15 +175,12 @@ export type Cartesian<L, R> =
   ? [...{ [K in keyof R]: K extends `${number}` ? [L, R[K]] : never }]
   : [L, R];
 
-export const median = <T extends number | bigint>(arr: T[], isSorted: boolean = false): T => {
+export function median(arr: number[], isSorted?: boolean): number;
+export function median(arr: bigint[], isSorted?: boolean): bigint;
+export function median(arr: (number | bigint)[], isSorted: boolean = false): number | bigint {
   if (arr.length === 0) throw new Error("Can't calculate median of empty array");
 
-  const sorted = isSorted
-    ? arr
-    : [...arr].sort((a, b) => {
-        // handle bigint and number
-        return a > b ? 1 : a < b ? -1 : 0;
-      });
+  const sorted = isSorted ? arr : [...arr].sort((a, b) => (a > b ? 1 : a < b ? -1 : 0)); // handle bigint and number
 
   const mid = Math.floor(sorted.length / 2);
 
@@ -195,10 +192,12 @@ export const median = <T extends number | bigint>(arr: T[], isSorted: boolean = 
   const right = sorted[mid]!;
 
   if (typeof left === "bigint" && typeof right === "bigint") {
-    return ((left + right) / 2n) as T;
-  } else if (typeof left === "number" && typeof right === "number") {
-    return ((left + right) / 2) as T;
-  } else {
-    throw new Error("Can't calculate median of array with mixed number and bigint");
+    return (left + right) / 2n;
   }
-};
+
+  if (typeof left === "number" && typeof right === "number") {
+    return (left + right) / 2;
+  }
+
+  throw new Error("Can't calculate median of array with mixed number and bigint");
+}
--- a/core/base/src/utils/misc.ts
+++ b/core/base/src/utils/misc.ts
@@ -25,3 +25,7 @@ export function throws(fn: () => any): boolean {
     return true;
   }
 }
+
+export function bound<T extends number | bigint>(value: T, min: T, max: T): T {
+  return min > value ? min : max < value ? max : value;
+}
--- a/core/base/src/utils/misc.ts
+++ b/core/base/src/utils/misc.ts
@@ -26,6 +26,14 @@ export function throws(fn: () => any): boolean {
   }
 }
 
-export function bound<T extends number | bigint>(value: T, min: T, max: T): T {
-  return min > value ? min : max < value ? max : value;
+export function bound(value: number, min: number, max: number): number;
+export function bound(value: bigint, min: bigint, max: bigint): bigint;
+export function bound(
+  value: number | bigint,
+  min: number | bigint,
+  max: number | bigint,
+): number | bigint {
+  if (value < min) return min;
+  if (value > max) return max;
+  return value;
 }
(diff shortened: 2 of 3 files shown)


platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,80 @@ export async function determinePriorityFee(
...
+        (a, b) => a.prioritizationFee - b.prioritizationFee,
+      );
+
+      // Calculate the median
+      const half = Math.floor(sortedFees.length / 2);
+      if (sortedFees.length % 2 === 0) {
+        fee = Math.floor(
+          (sortedFees[half - 1]!.prioritizationFee +
+            sortedFees[half]!.prioritizationFee) /
+            2,
```
abhidtu2014, line 531: nit:
```suggestion
          (sortedFees[half - 1]!.prioritizationFee + (sortedFees[half]!.prioritizationFee) >> 1)
```

abhidtu2014 (commented):
Added some nit comments

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,80 @@ export async function determinePriorityFee(
...
+    const recentPrioritizationFees =
+      response.result as RecentPrioritizationFees[];
+
+    if (recentPrioritizationFees.length > 0) {
+      const sortedFees = recentPrioritizationFees.sort(
+        (a, b) => a.prioritizationFee - b.prioritizationFee,
+      );
+
+      // Calculate the median
+      const half = Math.floor(sortedFees.length / 2);
```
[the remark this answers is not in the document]
kev1n-peters: We only subtract `1` from `half` when `sortedFees.length` is greater than `0` and even, so `half` won't ever be negative. I would prefer to keep `/ 2` instead of the right shift since it's clearer (it may not be obvious to someone what `>> 1` is doing)

abhidtu2014 (approved):

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+
+  // @ts-ignore
+  const rpcRequest = connection._rpcRequest;
+
+  const accounts = await getWritableAccounts(connection, transaction);
+
+  const args = [
+    accounts,
+    {
+      percentile: percentile * 10_000, // between 1 and 10_000
```
artursapek, line 504: i would validate the `percentile` input is a number between 0 and 1, and throw if it's not.

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+
+  if (recentPrioritizationFees.length > 0) {
+    const sortedFees = recentPrioritizationFees.sort(
+      (a, b) => a.prioritizationFee - b.prioritizationFee,
+    );
+
+    // Calculate the median
+    const half = Math.floor(sortedFees.length / 2);
+    if (sortedFees.length % 2 === 0) {
+      fee = Math.floor(
```
nonergodic, line 529: not much of a point using `Math.floor` here if we potentially multiply with `mulitple` after

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+        (sortedFees[half - 1]!.prioritizationFee +
+          sortedFees[half]!.prioritizationFee) /
+          2,
+      );
+    } else {
+      fee = sortedFees[half]!.prioritizationFee;
+    }
+
+    if (multiple > 0) {
```
nonergodic, line 538: why isn't it simply 1 by default? using 0 as a special value seems strange

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+      fee *= multiple;
+    }
+  }
+
+  // Bound the return value by the parameters passed
+  return Math.min(Math.max(fee, minPriorityFee), maxPriorityFee);
```
nonergodic, line 544:
```suggestion
  const rpcRequest = connection._rpcRequest;

  const accounts = await getWritableAccounts(connection, transaction);

  const args = [
    accounts,
    {
      percentile: percentile * 10_000, // between 1 and 10_000
    },
  ];

  const response = (await rpcRequest(
    'getRecentPrioritizationFees',
    args,
  )) as RpcResponse;

  if (response.error) {
    throw new Error(response.error);
  }

  const recentPrioritizationFees = (response.result as RecentPrioritizationFees[]).map(e => e.prioritizationFee);
  
  if (recentPrioritizationFees.length === 0)
    return minPriorityFee;

  //TODO: move this to e.g. core/base/utils/array.ts and consider implementing for bigint too
  const median = (arr: number[], isSorted: boolean = false) => {
    if (arr.length === 0)
      throw new Error("Can't calculate median of empty array");

    const sorted = isSorted ? arr : [...arr].sort((a,b) => a - b);
    const mid = Math.floor(sorted.length / 2);
    return sorted.length % 2 === 0
      ? (sorted[mid-1] + sorted[mid]) / 2
      : sorted[mid];
  };

  const unboundedFee = Math.Floor(median(recentPrioritizationFees) * (multiple > 0 ? multiple : 1));
  
  //TODO: move this to some utility too and also consider also implementing it for bigint
  const bound = (val: number, lower: number, upper: number): T =>
    Math.min(Math.max(val, lower), upper);

  return bound(unboundedFee, minPriorityFee, maxPriorityFee);
```
See related questions that should figure into this suggestion.

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+    'getRecentPrioritizationFees',
+    args,
+  )) as RpcResponse;
+
+  if (response.error) {
+    // Allow the error to propagate so the caller can handle it and fall-back if needed
+    throw new Error(response.error);
+  }
+
+  const recentPrioritizationFees =
```
nonergodic, line 518: Why not `map` to a number array immediately? the array of objects with `prioritizationFee` fields is purely a nuissance.

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,78 @@ export async function determinePriorityFee(
...
+export async function determinePriorityFeeTritonOne(
+  connection: Connection,
+  transaction: Transaction | VersionedTransaction,
+  percentile: number = DEFAULT_PRIORITY_FEE_PERCENTILE,
+  multiple: number = DEFAULT_PERCENTILE_MULTIPLE,
+  minPriorityFee: number = DEFAULT_MIN_PRIORITY_FEE,
+  maxPriorityFee: number = DEFAULT_MAX_PRIORITY_FEE,
+): Promise<number> {
+  // Start with min fee
+  let fee = minPriorityFee;
```
nonergodic, line 494: Prefer const vars and early returns over mutating variables when possible (see suggestion).

platforms/solana/src/signer.ts
kev1n-peters: good idea

platforms/solana/src/signer.ts
kev1n-peters: agreed, addressed in refactor

platforms/solana/src/signer.ts
kev1n-peters: addressed in refactor

platforms/solana/src/signer.ts
kev1n-peters: agree, addressed in refactor

platforms/solana/src/signer.ts
kev1n-peters: I added the `median` and `bound` utility functions in the latest commit. It was kind of awkward supporting number and bigint (need to check the types in some cases), but it works

core/base/src/utils/array.ts
```diff
@@ -174,3 +174,31 @@ export type Cartesian<L, R> =
   : R extends RoArray
   ? [...{ [K in keyof R]: K extends `${number}` ? [L, R[K]] : never }]
   : [L, R];
+
+export const median = <T extends number | bigint>(arr: T[], isSorted: boolean = false): T => {
```
nonergodic, line 178: I think overloading might be the better approach here because it explicitly disallows mixed arrays, i.e.:
```
function median(arr: number[], isSorted: boolean = false): number;
function median(arr: bigint[], isSorted: boolean = false): bigint;
function median(arr: (number | bigint)[], isSorted: boolean = false): number | bigint {
  // impl goes here
}
```
This allows you to skip the error checking at the end and simply condition on the type of one of the elements.

core/base/src/utils/misc.ts
```diff
@@ -25,3 +25,7 @@ export function throws(fn: () => any): boolean {
     return true;
   }
 }
+
+export function bound<T extends number | bigint>(value: T, min: T, max: T): T {
```
nonergodic, line 29: Same logic re overloading here, I'd say.
The problem with the generic argument is that it allows for constructions like:
```
function foo(val: number): number | bigint {
  return val % 2 === 0 ? val : BigInt(val);
}

//T is deduced to be number | bigint and will do mixed comparisons internally
bound(foo(1), foo(2), foo(3));
```

platforms/solana/src/signer.ts
```diff
@@ -461,6 +472,62 @@ export async function determinePriorityFee(
...
+export async function determinePriorityFeeTritonOne(
+  connection: Connection,
+  transaction: Transaction | VersionedTransaction,
+  percentile: number = DEFAULT_PRIORITY_FEE_PERCENTILE,
+  multiple: number = DEFAULT_PERCENTILE_MULTIPLE,
+  minPriorityFee: number = DEFAULT_MIN_PRIORITY_FEE,
+  maxPriorityFee: number = DEFAULT_MAX_PRIORITY_FEE,
+): Promise<number> {
+  if (percentile <= 0 || percentile > 1) {
+    throw new Error('percentile must be between 0 and 1');
```
nonergodic, line 494: This check is inconsistent with the old comment (e.g. percentile = 1e-5 will result in a value of 0.1 in `args` decidedly not between 1 and 10k).

core/base/src/utils/array.ts
```diff
@@ -174,3 +174,31 @@ export type Cartesian<L, R> =
...
+  if (arr.length === 0) throw new Error("Can't calculate median of empty array");
+
+  const sorted = isSorted
+    ? arr
+    : [...arr].sort((a, b) => {
+        // handle bigint and number
+        return a > b ? 1 : a < b ? -1 : 0;
+      });
```
nonergodic, line 186:
```suggestion
    : [...arr].sort((a, b) => a > b ? 1 : a < b ? -1 : 0); // handle bigint and number
```

core/base/src/utils/array.ts
kev1n-peters: I still get the `Operator '+' cannot be applied to types 'number | bigint' and 'number | bigint'.ts(2365)` error when adding these overloads and checking the type of just one element unfortunately

core/base/src/utils/misc.ts
kev1n-peters: makes sense

core/base/src/utils/array.ts
nonergodic: I was writing it sloppily.
The actual type of `arr` should be `RoArray<number> | RoArray<bigint>` then it should (hopefully) narrow correctly (or worst case you narrow manually yourself)).

artursapek (approved):

PR #787, opened 2025-01-21
Added UnattestedTokenId type

kev1n-peters:
Represents a token that does not exist yet

### DIFF
--- a/platforms/evm/src/platform.ts
+++ b/platforms/evm/src/platform.ts
@@ -105,7 +105,7 @@ export class EvmPlatform<N extends Network>
       rpc,
       new EvmAddress(token).toString(),
     );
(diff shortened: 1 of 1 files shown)


artursapek (dismissed):

connect/src/routes/request.ts
```diff
@@ -86,15 +86,21 @@ export class RouteTransferRequest<N extends Network> {
     params: {
       source: TokenId<FC>;
       destination: TokenId<TC>;
+      sourceDecimals?: number;
```
kev1n-peters, line 89: We can pass in the decimals as an optimization (so we don't have to fetch if already known) and also when the token doesn't exist yet (e.g. unattested destination token) which would throw an error.

artursapek (approved):

PR #792, opened 2025-02-06
Add fake Hyperliquid chain

artursapek:


kev1n-peters:
This is a hack to get the Hyperliquid route working in Connect.

kev1n-peters (approved):

PR #822, opened 2025-03-20
Changed deprecated public Ankr RPC endpoints

kev1n-peters:
Ankr RPC endpoints now require an API key


fergarrui (dismissed):

platforms/evm/__tests__/integration/tokenBridge.test.ts
```diff
@@ -82,7 +81,7 @@ afterEach(async () => {
   nockBack.setMode('wild');
 });
 
-const network = DEFAULT_NETWORK;
```
kev1n-peters, line 85: These tests were actually failing when run locally since the environment would default to TESTNET. The mocked response data is for Mainnet, so this should be set to Mainnet...

platforms/evm/__tests__/unit/platform.test.ts
```diff
@@ -35,7 +34,7 @@ const EVM_CHAINS = chains.filter(
   (c) => chainToPlatform(c) === EvmPlatform._platform,
 ) as EvmChains[];
 
-const network = DEFAULT_NETWORK;
```
kev1n-peters, line 38: Same as above

fergarrui (approved):

PR #830, opened 2025-04-11
Add expiration time to `Quote`

artursapek:
- expose how long a quote is good for
- set to 5 minutes for relayed routes
- set to 24 hours for manual routes (which technically never expire)


core/base/src/utils/time.ts
```diff
@@ -0,0 +1,6 @@
+export const expiration = (hours: number = 0, minutes: number = 0, seconds: number = 0): Date => {
+  return (new Date(
+    new Date().valueOf() +
```
kev1n-peters, line 3: nit: Can we use `Date.now()` instead which expresses intent clearer?

kev1n-peters (approved):

PR #849, opened 2025-05-27
added sonic and linea usdc

kev1n-peters:


connect/src/protocols/cctp/cctpTransfer.ts
```diff
@@ -582,9 +582,13 @@ export namespace CircleTransfer {
     dstChain: ChainContext<N, Chain>,
     transfer: Omit<CircleTransferDetails, "from" | "to">,
   ): Promise<TransferQuote> {
+    if (!circle.isCircleChain(dstChain.network, dstChain.chain))
```
kev1n-peters, line 585: Before we would return a valid quote if USDC existed on both of the chains, but now we need to make sure that both chains are CCTP V1 domains (Sonic and Linea only support V2 which the Wormhole SDK doesn't support)

PR #854, opened 2025-06-02
add Converge testnet

aapclark:
- add Converge to chain constants

- add RPC value for Converge testnet


tonyjin (changes requested):
Wrong RPC URL

tonyjin (approved):

PR #857, opened 2025-06-05
add mezo mainnet

aapclark:
- add mezo mainnet core bridge

- add mezo mainnet, chainId, rpc url, and explorer

- add mezo testnet testnet


tonyjin (approved):
lgtm. RPC + chain ID match https://mezo.org/docs/users/getting-started/connect, core contract address looks correct

PR #856, opened 2025-06-05
removed Terra, Oasis, Aurora, Karura, Acala, and Xpla chains

kev1n-peters:

PR #865, opened 2025-06-12
partially revert #856

aadam-10:


platforms/evm/src/weth.ts
```diff
@@ -10,6 +10,7 @@ export const WETH_CONTRACTS: Record<Network, Partial<Record<EvmChains, string>>>
     "Celo": "0x0000000000000000000000000000000000000000",
     "Ethereum": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
     "Fantom": "0x21be370D5312f44cB42ce377BC9b8a0cEF1A4C83",
+    "Karura": "0x0000000000000000000000000000000000000000",
```
bruce-riley, line 13: Just to confirm, there is no Acala here?

aadam-10: In the original PR Acala wasn't mentioned: https://github.com/wormhole-foundation/wormhole-sdk-ts/pull/856/files#diff-d820f6f4ebaaac7579496093218325313eefd691df6daab70bcb14fee7cc6261L14

aadam-10: added, good catch

PR #871, opened 2025-07-07
feat: add ExecutorTokenBridge route with gas dropoff and referrer fee…

kev1n-peters:
… support


connect/src/executor-api.ts
```diff
@@ -0,0 +1,119 @@
+import { type Chain, toChainId } from "@wormhole-foundation/sdk-base";
+import type { SignedQuote, TxHash } from "@wormhole-foundation/sdk-definitions";
+import axios from "axios";
+
+export enum RelayStatus {
```
evan-gray, line 5: can these enums and types go in definitions instead of connect, please? 🙏

kev1n-peters: Moved these to the definitions SDK. Decided to keep the API fetch functions in the connect SDK for now. We could potentially move those to the definitions SDK and use `fetch` instead of `axios`, although I'm not sure if fetch functions should belong in a "definitions" SDK 🤔

kev1n-peters: Ultimately decided to move the fetch functions into the definitions SDK using fetch() instead of axios so that they can be used elsewhere from the light-weight definitions package.

connect/src/protocols/tokenBridge/tokenTransfer.ts
```diff
@@ -902,4 +1165,22 @@ export namespace TokenTransfer {
 
     return _transfer;
   }
+
+  export function calculateReferrerFee(
```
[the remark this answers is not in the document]
kev1n-peters: I think that it's okay if the integer division rounds down to 0 here. An alternative is returning 1 base unit instead of 0, but I think returning 0 is the simpler approach

PR #880, opened 2025-07-24
Add `sender` to `RouteTransferRequest`

artursapek:


priyajeet (approved):

connect/src/routes/request.ts
```diff
@@ -23,11 +24,13 @@ export class RouteTransferRequest<N extends Network> {
     source: TokenDetails,
     destination: TokenDetails,
     recipient?: ChainAddress,
+    sender?: ChainAddress,
```
kev1n-peters, line 27: Should sender come before recipient here?

PR #885, opened 2025-07-31
Handle special address that represents ETH

artursapek:
Covalent started returning the address `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` to represent the wallet's native gas balance. The code downstream was treating this as an ERC-20 and failing to load decimals, since it's not a valid token.

### DIFF
--- a/platforms/evm/src/indexer.ts
+++ b/platforms/evm/src/indexer.ts
@@ -52,7 +52,13 @@ export class GoldRushClient {
 
     const bals: Balances = {};
     for (let item of data.items) {
-      bals[item.contract_address] = BigInt(item.balance);
+      let addr = item.contract_address;
+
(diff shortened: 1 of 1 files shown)


platforms/evm/src/indexer.ts
```diff
@@ -52,7 +52,13 @@ export class GoldRushClient {
 
     const bals: Balances = {};
     for (let item of data.items) {
-      bals[item.contract_address] = BigInt(item.balance);
+      let addr = item.contract_address;
+
+      if (item.contract_address.toLowerCase() === '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee') {
```
emreboga, line 57: Maybe comment here about why?

artursapek: I think this code is pretty self-explanatory

PR #899, opened 2025-08-19
fix: use goldrush as a fallback

kyleleighton:
- Added a 3-second timeout to GoldRush API calls
- Added Alchemy as a fallback (5-second timeout) before falling back to individual RPC calls

The flow now
Try GoldRush (3s max)
If that fails/times out → try Alchemy (5s max)
If both fail → fall back to individual RPC calls (existing behavior)
Worst case is now ~8 seconds instead of 30+ seconds of hanging

When combined with an addition of the [alchemy key in portal](https://github.com/wormholelabs-xyz/portal-nextjs/blob/d144592220251ebc14b8cd442f6ed19d1c20fa6e/src/hooks/useConnectConfig.ts#L199-L201) this will ensure that fallback pattern and fixes

### DIFF
--- a/platforms/evm/src/platform.ts
+++ b/platforms/evm/src/platform.ts
@@ -157,7 +157,7 @@ export class EvmPlatform<N extends Network>
 
       try {
         if (!provider.supportsChain(network, chain)) {
-          console.error(
+          console.info(
             `Network=${network} Chain=${chain} not supported by ${name} indexer API`,
           );
           return null;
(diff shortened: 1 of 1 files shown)


platforms/evm/src/platform.ts
```diff
@@ -135,39 +140,79 @@ export class EvmPlatform<N extends Network>
...
+      timeoutMs: number,
+    ): Promise<Balances | null> => {
+      if (!provider) {
+        return null;
       }
-      if (indexers.alchemy) {
+
+      try {
+        if (!provider.supportsChain(network, chain)) {
+          console.error(
```
priyajeet, line 160: Probably info/log or nothing

priyajeet (approved):

PR #906, opened 2025-09-02
Fix Aptos CircleBridge MessageSent event parsing

DBosley:
## Summary
- Fixes "No MessageSent event found" error when bridging from Aptos using CircleBridge
- Adds retry logic to handle event availability timing issues

## Problem
Transactions succeed on-chain but SDK throws error because events aren't immediately available when parseTransactionDetails is called.

## Solution
- Add 10 retry attempts with 2s delay between attempts
- Preserve address normalization for compatibility
- Handle both missing events and RPC errors gracefully

## Test
Verified with transaction: 0xd4dcf554efac49d3bec498b4134c1f7c8eab75c02bb595cd046d86a6ff96917f


platforms/aptos/protocols/cctp/src/circleBridge.ts
```diff
@@ -123,15 +125,47 @@ export class AptosCircleBridge<N extends Network, C extends AptosChains>
...
 
   async parseTransactionDetails(digest: string): Promise<CircleTransferMessage> {
-    const tx = await this.provider.getTransactionByHash({ transactionHash: digest });
-
-    const messageTransmitterId = this.messageTransmitterId.replace(/^0x0+/, "0x"); // remove any leading zeros to match event
-    const circleMessageSentEvent = (tx as UserTransactionResponse).events?.find(
-      (e) => e.type === `${messageTransmitterId}::message_transmitter::MessageSent`,
-    );
-
+    const retries = 10;
```
kev1n-peters, line 128: Can we do 5 retries and sleep for 1 second between them?

kev1n-peters (approved):

PR #908, opened 2025-09-05
feat: add explorer url to RelayFailedError

kev1n-peters:
Useful for UI to show recovery link when relay fails


connect/src/routes/types.ts
```diff
@@ -119,8 +119,17 @@ export class UnavailableError extends Error {
   }
 }
 
+export interface RelayExplorer {
+  url: string;
+  explorerName: string;
```
priyajeet, line 124: just `name`? or instead `exporerUrl`

priyajeet (approved):

PR #924, opened 2025-09-25
feat: creditcoin executor address

douglasgalico:


core/base/src/constants/contracts/executor.ts
```diff
@@ -12,6 +12,7 @@ export const executorContracts = [[
     ["Berachain",       "0x0Dd7a5a32311b8D87A615Cc7f079B632D3d5e2D3"],
     ["Bsc",             "0xeC8cCCD058DbF28e5D002869Aa9aFa3992bf4ee0"],
     ["Celo",            "0xe6Ea5087c6860B94Cf098a403506262D8F28cF05"],
+    ["CreditCoin",      "0xd2e420188f17607Aa6344ee19c3e76Cf86CA7BDe"],
```
kev1n-peters, line 15: Are there token bridge executor addresses yet? Those would also need to be added in the tokenBridgeExecutor contracts file

priyajeet (approved):

PR #966, opened 2025-12-19
fix: add MegaETH RPC urls

aapclark:


core/base/src/constants/rpc.ts
```diff
@@ -99,6 +100,7 @@ const rpcConfig = [[
     ["CreditCoin",      "https://rpc.cc3-testnet.creditcoin.network"],
     ["Stacks",          "https://api.testnet.hiro.so"],
     ["Moca",            "https://testnet-rpc.mocachain.org/"],
+    ["MegaETH",         "https://mainnet.megaeth.com/rpc"],
```
kev1n-peters: Mainnet and testnet use the same RPC?

PR #997, opened 2026-04-02
feat: support executor callback overrides on WormholeConfig

kev1n-peters:
Allow users to pass a getCapabilities callback when constructing the Wormhole SDK object so every route can use it without per-route config.

## Summary by CodeRabbit

* **New Features**
  * Added optional custom executor capabilities fetcher so users can supply their own async capability retrieval.
* **Behavior Changes**
  * Configuration partials now preserve function-typed properties (function types remain intact in overrides), improving type accuracy and developer ergonomics.


kev1n-peters:
@claude review this pr and make sure no bugs were introduced, and the getCapabilities callback will work correctly when passed in

PR #1013, opened 2026-06-08
feat: migrate Sui to @mysten/sui v2 (gRPC) and make the SDK ESM-only

kev1n-peters:
Migrate the Sui platform from @mysten/sui ^1.37 (JSON-RPC) to ^2.17 using the gRPC
client (SuiGrpcClient + client.core), and drop CommonJS across the whole repo in
favor of ESM-only.

@mysten/sui v2 is ESM-only, which forced the Sui packages to ESM-only and created a
dual-package hazard: a CommonJS require of the aggregator loaded a second ESM copy of
sdk-connect, breaking cross-platform address registration. Going ESM-only everywhere
removes the hazard and keeps builds consistent. ESM, bundler, and Node >=20.19
(import / require(esm)) consumers are unaffected; raw CommonJS require() is no longer
supported (breaking change).

Sui gRPC migration:
- SuiGrpcClient at the public fullnode gRPC endpoint; flat object.json reads; core
  methods (listCoins/listOwnedObjects/listDynamicFields/getDynamicField, getBalance/
  listBalances, getCoinMetadata, simulateTransaction, executeTransaction, getTransaction)
- BCS-encoded dynamic-field names; getLatestBlock via ledgerService; chainFromRpc via
  core.getChainIdentifier (base58 digest -> first-4-bytes hex)
- create-wrapped discovery via owned objects (gRPC-native, no GraphQL)

ESM-only:
- 31 packages -> ESM-only exports/main/types, drop build:cjs, engines >=20.19
- remove all tsconfig.cjs.json + connect index.d.cts/.d.mts; sdk aggregator ESM-only

Tests:
- jest runs in ESM mode (extensionsToTreatAsEsm + ts-jest useESM,
  --experimental-vm-modules, transpile-only); per-package configs
- import { jest } from '@jest/globals' where the mock API is used;
  unstable_mockModule + dynamic import for solana platform & connect axios mocks;
  import.meta.url __dirname shims for nock integration tests
- Sui live-RPC integration tests for data-access, core bridge, and token-bridge methods

Clean turbo build 35/35; turbo lint 35/35; all workspace test suites pass.


platforms/sui/src/platform.ts
```diff
@@ -83,89 +83,93 @@ export class SuiPlatform<N extends Network>
...
   }
 
-  static async sendWait(chain: Chain, rpc: SuiClient, stxns: SignedTx[]): Promise<TxHash[]> {
+  static async sendWait(chain: Chain, rpc: SuiGrpcClient, stxns: SignedTx[]): Promise<TxHash[]> {
     const txhashes = [];
     for (const stxn of stxns) {
-      const pendingTx = await rpc.executeTransactionBlock(stxn);
-      await rpc.waitForTransaction({ digest: pendingTx.digest });
-      txhashes.push(pendingTx.digest);
+      const result = await rpc.executeTransaction(stxn as any);
```
bengtlofgren: Curse whoever decided signed_transactions -> `stxns` XDD - but also do we need the `as any` cast here?

bengtlofgren: Looks like `stxn` isnt even typed to begin with - feels like it would be nice to type it even if it only is used in this file? Just so I know what im looking at

kev1n-peters: We can remove it, it just adds noise

platforms/sui/src/platform.ts
```diff
@@ -83,89 +83,93 @@ export class SuiPlatform<N extends Network>
...
 
-  static async sendWait(chain: Chain, rpc: SuiClient, stxns: SignedTx[]): Promise<TxHash[]> {
+  static async sendWait(chain: Chain, rpc: SuiGrpcClient, stxns: SignedTx[]): Promise<TxHash[]> {
     const txhashes = [];
     for (const stxn of stxns) {
-      const pendingTx = await rpc.executeTransactionBlock(stxn);
-      await rpc.waitForTransaction({ digest: pendingTx.digest });
-      txhashes.push(pendingTx.digest);
+      const result = await rpc.executeTransaction(stxn);
+      const digest = (result.Transaction ?? result.FailedTransaction)!.digest;
```
bengtlofgren: Is this intentional to take the digest irrespective of if it was successful tx or not? No error logged or anything?

kev1n-peters: We should probably throw if the tx failed on-chain

connect/package.json
```diff
@@ -13,160 +13,67 @@
...
-        "./dist/cjs/circle-api.d.ts"
+        "./dist/esm/circle-api.d.ts"
       ],
       "whscan-api": [
-        "./dist/cjs/whscan-api.d.ts"
+        "./dist/esm/whscan-api.d.ts"
       ],
       "*": [
         "./index.d.cts"
       ]
```
bengtlofgren, line 72: Claude pointed out that this file was deleted in this pr and I think he's right

platforms/sui/protocols/tokenBridge/src/utils.ts
```diff
@@ -1,95 +1,90 @@
...
+    const { object } = await provider.getObject({
+      objectId: dynamicField.fieldId,
+      include: { json: true },
+    });
+    if (!object || !object.json) return null;
+    return { type: object.type, json: object.json as Record<string, any> };
+  } catch {
+    // dynamic field not found → token not registered
+    return null;
   }
```
bengtlofgren, line 58: I guess we return null on anything unexpected atm, but it seems a bit meh. Up to you if you want to make more specific or not

connect/package.json
kev1n-peters: removing

platforms/sui/protocols/tokenBridge/src/utils.ts
kev1n-peters: I think we should just leave it for now since it still preserves the fail closed behavior from before

douglasgalico (approved):

PR #1029, opened 2026-07-17
fix: use coinWithBalance for Sui CCTP USDC coin resolution

Stanoja:
## Summary

- Replace manual coin fetching/merging/splitting with `coinWithBalance` helper from `@mysten/sui/transactions`
- Fixes "No USDC in wallet" errors caused by gRPC `SuiGrpcClient.listCoins` not returning certain coin objects after the `@mysten/sui` v2 upgrade

## Context

The `@mysten/sui` v2 upgrade switched to gRPC, which has a bug where certain coin objects (e.g., those produced by `accumulator_settlement`) are invisible to `SuiGrpcClient.listCoins` but still exist on-chain.

See: https://github.com/MystenLabs/sui/issues/26875

In order for our application to work, I had to patch this exact change, and it's been running as expected for the past 10 days.

### DIFF
--- a/platforms/sui/protocols/cctp/src/circleBridge.ts
+++ b/platforms/sui/protocols/cctp/src/circleBridge.ts
@@ -1,5 +1,5 @@
 import type { SuiGrpcClient } from "@mysten/sui/grpc";
-import { Transaction } from "@mysten/sui/transactions";
+import { coinWithBalance, Transaction } from "@mysten/sui/transactions";
 import { SuiPlatform, type SuiChains, SuiUnsignedTransaction } from "@wormhole-foundation/sdk-sui";
 import type {
   AccountAddress,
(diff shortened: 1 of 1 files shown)


Stanoja:
@johnsaigle May I know why is this closed?

johnsaigle:
This PR modifies the semantics of how the Sui coin data gets built in the transaction. There's not a need on our side to update this flow so we're not looking to upstream this change.

Stanoja:
> This PR modifies the semantics of how the Sui coin data gets built in the transaction. There's not a need on our side to update this flow so we're not looking to upstream this change.

But the gRPC `listCoins` does not return the same result as the JSONRPC `getCoins`.
So your suggestion is to wait for the Sui team to address the issue of gRPC's `listCoins`?

Meahwhile, I've patched this locally for the past week or so, seems to be working fine.

PR #1042, opened 2026-09-13
fix: populate destination txs in receipts once the API has them

### LINKED ISSUE #659: connect: receipt destination tx hash not always set
There are some cases where the destination tx hash won't be set even after the tx is completed. We should try as best we can to find the destination tx hash (check wormholescan or fall-back to fetching on-chain data). This is applicable to every route.

MarkFeder:
Fixes #659: receipts could end up with a missing (or dropped) destination tx hash.

- CCTP: `CircleTransfer.track` overwrote the hash found via WormholeScan with an empty list once the transfer completed on-chain; the known hash is now preserved and backfilled with a last-chance lookup when the backend was slow.
- Token bridge, automatic Portico and tBTC: same last-chance lookup fills `destinationTxs` at finalization; Portico/tBTC previously never set it.

Adds a single-shot `getDestinationTx` helper on the WormholeScan API client plus tests covering both the happy path and the delayed-backend case.


MarkFeder:
Hey @kev1n-peters @aadam-10, could you take a look? Receipts now keep the destination tx hash found via WormholeScan (instead of overwriting it with an empty list at finalization) and backfill it with a last-chance lookup when the backend was slow (fixes #659).

PR #1044, opened 2026-09-14
fix: avoid duplicate completion in checkAndCompleteTransfer

### LINKED ISSUE #1039: checkAndCompleteTransfer can retry completion with a stale receipt
`checkAndCompleteTransfer()` appears to drop the updated receipt returned by `route.complete()` / `route.finalize()`. Because the helper then recurses with the old receipt, a manual route can keep seeing the transfer as `Attested` until destination completion is externally observed, and can submit another redeem/complete transaction for the same attestation.

I reviewed this on commit `343e3d00a5c6`.

### Where the state is lost

In `connect/src/routes/common.ts`, the async generator updates the local `receipt` while tracking. But after a completion/finalization transaction is submitted, the returned receipt is assigned to a variable named like txids and is not written back to `receipt`:

```ts
// connect/src/routes/common.ts:15-60
export async function checkAndCompleteTransfer<N extends Network>(
  route: Route<N>,
  receipt: Receipt,
  destinationSigner?: Signer<N>,
  timeout: number = 120 * 1000,
  // byo logger but im dumping to console rn 🙃
  log: typeof console.log = console.log,
) {
  const start = Date.now();
  log("Checking transfer state...");

  // overwrite receipt var as we receive updates, will return when it's complete
  // but can be called again if the destination is not finalized
  // this construct is to drain an async generator and return the last value
  for await (receipt of route.track(receipt, 120 * 1000)) {
    log("Current Transfer State: ", TransferState[receipt.state]);
  }

  // gucci
  if (isCompleted(receipt)) return receipt;

  // if the route is one we need to complete, do it
  if (isManual(route) && isAttested(receipt) && destinationSigner) {
    log("Completing transfer...");
    const completedTxids = await route.complete(destinationSigner, receipt);
    log("Completed transfer with txids: ", completedTxids);
  }

  // if the route is one we need to finalize, do it
  if (isFinalizable(route) && isRedeemed(receipt) && destinationSigner) {
    log("Finalizing transfer...");
    const completedTxids = await route.finalize(destinationSigner, receipt);
    log("Finalized transfer with txids: ", completedTxids);
  }

  const leftover = timeout - (Date.now() - start);
  // do we still have time?
  if (leftover > 0) {
    // give it a second, computers need to rest sometimes
    const wait = 2 * 1000;
    log(`Transfer not complete, trying again in a ${wait}ms...`);
    await new Promise((resolve) => setTimeout(resolve, wait));
    return checkAndCompleteTransfer(route, receipt, destinationSigner, leftover);
  }

  return receipt;
}
```

### `complete()` returns an updated receipt, not txids

The abstract route type declares that manual `complete()` returns `Promise<R>`:

```ts
// connect/src/routes/route.ts:127-136
export abstract class ManualRoute<
  N extends Network,
  OP extends Options = Options,
  VP extends ValidatedTransferParams<OP> = ValidatedTransferParams<OP>,
  R extends Receipt = Receipt,
> extends Route<N, OP, VP, R> {
  static NATIVE_GAS_DROPOFF_SUPPORTED = false;
  static IS_AUTOMATIC = false;
  public abstract complete(sender: Signer, receipt: R): Promise<R>;
  public abstract resume(tx: TransactionId): Promise<R>;
}
```

The manual token bridge implementation returns a receipt advanced to `DestinationInitiated` with `destinationTxs`:

```ts
// connect/src/routes/tokenBridge/manual.ts:184-199
async complete(signer: Signer, receipt: R): Promise<R> {
  if (!isAttested(receipt))
    throw new Error("The source must be finalized in order to complete the transfer");
  const toChain = this.wh.getChain(receipt.to);
  const dstTxIds = await TokenTransfer.redeem<N>(
    toChain,
    receipt.attestation.attestation as TokenTransfer.VAA,
    signer,
  );

  return {
    ...receipt,
    state: TransferState.DestinationInitiated,
    destinationTxs: dstTxIds,
  };
}
```

CCTP manual does the same:

```ts
// connect/src/routes/cctp/manual.ts:171-194
async complete(signer: Signer, receipt: R): Promise<R> {
  if (!isAttested(receipt))
    throw new Error("The source must be finalized in order to complete the transfer");

  const { id, attestation: att } = receipt.attestation;
  if (CircleBridge.isCircleAttestation(att)) {
    const { message, attestation } = att;
    if (!attestation) throw new Error(`No Circle attestation for ${id}`);

    const toChain = this.wh.getChain(receipt.to);
    const cb = await toChain.getCircleBridge();
    const sender = Wormhole.parseAddress(signer.chain(), signer.address());
    const xfer = cb.redeem(sender, message, attestation);
    const dstTxids = await signSendWait<N, Chain>(toChain, xfer, signer);
    return {
      ...receipt,
      state: TransferState.DestinationInitiated,
      destinationTxs: dstTxids,
    };
  } else {
    //
    return receipt;
  }
}
```

### Why this matters

A typical sequence is:

1. `route.track()` returns an `Attested` manual receipt.
2. `checkAndCompleteTransfer()` calls `route.complete(destinationSigner, receipt)`.
3. The route submits the destination redeem transaction and returns a `DestinationInitiated` receipt.
4. `checkAndCompleteTransfer()` discards that returned receipt and recurses with the old `Attested` receipt.
5. If the next tracking pass does not yet observe destination finality, the helper can satisfy `isManual(route) && isAttested(receipt)` again and submit another completion transaction for the same attestation.

Depending on destination-chain timing and route behavior, this can produce duplicate redeem attempts, extra gas spend, confusing logs (`Completed transfer with txids` prints a receipt), or a later revert for an already-redeemed attestation. The core contracts may still reject double redemption, but the SDK helper should preserve the lifecycle state it just advanced.

### Suggested fix

Assign the returned receipt back to `receipt`:

```ts
if (isManual(route) && isAttested(receipt) && destinationSigner) {
  log("Completing transfer...");
  receipt = await route.complete(destinationSigner, receipt);
  log("Completed transfer with txids: ", receipt.destinationTxs);
}

if (isFinalizable(route) && isRedeemed(receipt) && destinationSigner) {
  log("Finalizing transfer...");
  receipt = await route.finalize(destinationSigner, receipt);
  log("Finalized transfer with txids: ", receipt.destinationTxs);
}
```

It may also be worth passing the caller-provided `timeout`/remaining budget into `route.track()` instead of the hard-coded `120 * 1000` at line 29, but the state overwrite above is the lifecycle issue that can trigger repeated completion attempts.

MarkFeder:
Fixes #1039.

`checkAndCompleteTransfer` dropped the receipt returned by `route.complete()` / `route.finalize()` and recursed with the stale one, so a manual route whose destination completion was not yet observable could submit a second redeem for the same attestation. It also mishandled its own timeout budget.

- Write the returned receipt back before the tracking retry, so the lifecycle state it just advanced is preserved, and log its `destinationTxs`
- Charge the wait between attempts to the remaining budget. It was computed before the wait and passed back as the full timeout, so the wait was never spent and the loop could run many times past the caller's timeout
- Track with the caller's `timeout` instead of a hard-coded `120 * 1000`
- Adds `connect/__tests__/checkAndCompleteTransfer.test.ts`: a stub route whose `track` has not yet observed destination finality. Without the state write-back `complete` is called twice; without the budget fix the retry loop runs away — the stub caps `track` calls so these fail with a clear error rather than hanging the suite

Verification: typecheck clean, connect suite 40/40.


MarkFeder:
Hey @kev1n-peters @aadam-10, could you take a look? `checkAndCompleteTransfer` dropped the receipt returned by `complete()`/`finalize()` and retried with the stale one — a manual route could redeem the same attestation twice. Small diff plus a stub-route regression test (fixes #1039).

